PR #36, opened 2022-08-09
ipfs media uploader

Geimaj:
Find processed tracks where `lossyAudioIPFSHash` is null and `lossyAudioURL` is set, or tracks where `lossyArtworkIPFSHash` is null and `lossyArtworkURL` is set and no error exists for that URL in the `ipfsFiles` table, then upload the files to IPFS and update the record with the IPFS hash and insert the url and CID into the `ipfsFiles` table. 

Uses local IPFS node, expects this to be running locally on the default port. 
New .env variables are required:

IPFS_NODE_MULTIADDRESS=address

You can get the address by running `ipfs id` on the ipfs node and copying one with the public ip


src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,51 @@
...
+import _ from 'lodash'
+
+import { Table } from '../../db/db';
+import { artworkNotOnIpfs, audioNotOnIpfs } from '../../triggers/ipfs';
+import { Clients, Processor } from '../../types/processor';
+import { ProcessedTrack } from '../../types/track';
+
+function processorFunction(sourceField: 'lossyAudioURL' | 'lossyArtworkURL', replaceField: 'lossyAudioIPFSHash' | 'lossyArtworkIPFSHash') {
+  return async (newTriggerItems: any, clients: Clients) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Geimaj, line 12: comment required because TS doesn't recognise the optional field `processedTrack.lossyAudioURL`. @musnit do you know how to fix this?

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,51 @@
...
+    // @ts-ignore
+    const tracksByFilePath = _.keyBy(newTriggerItems, (track: ProcessedTrack) => new URL(track[sourceField]).pathname.split('/').pop());
+      
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const sources = newTriggerItems.map((track: ProcessedTrack) => {return urlSource(track[sourceField]!)} );
```
Geimaj, line 18: same TS problem as above

src/clients/ipfs.ts
```diff
@@ -40,10 +43,19 @@ export const extractHashFromURL = (urlString: string) => {
 }
 
 const init = async () => {
+
+  const ETH_INFURA_IPFS_ID = process.env.INFURA_IPFS_ID
```
musnit, line 47: need to add to .env.example?

also, we should try make this based on a generic IPFS node rather than infura specific, so i'd use 2 env vars like:
 - IPFS_NODE_RPC_URL="https://ipfs.infura.io:5001/api/v0"
 - IPFS_NODE_AUTH_HEADER="Basic id:secret"

(with auth being optional if it's not needed)

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,51 @@
...
+import { urlSource } from 'ipfs-http-client';
+import _ from 'lodash'
+
+import { Table } from '../../db/db';
+import { artworkNotOnIpfs, audioNotOnIpfs } from '../../triggers/ipfs';
+import { Clients, Processor } from '../../types/processor';
+import { ProcessedTrack } from '../../types/track';
+
+function processorFunction(sourceField: 'lossyAudioURL' | 'lossyArtworkURL', replaceField: 'lossyAudioIPFSHash' | 'lossyArtworkIPFSHash') {
+  return async (newTriggerItems: any, clients: Clients) => {
```
musnit, line 11: shouldn't we just use a type here instead of any and ts-ignore?

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,51 @@
...
+    // @ts-ignore
+    const tracksByFilePath = _.keyBy(newTriggerItems, (track: ProcessedTrack) => new URL(track[sourceField]).pathname.split('/').pop());
+      
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const sources = newTriggerItems.map((track: ProcessedTrack) => {return urlSource(track[sourceField]!)} );
+    const allFiles = await clients.ipfs.client.addAll(sources, {
+      progress: (bytes, path) => console.log(`added ${bytes} bytes of ${path}`),
+      fileImportConcurrency: 10,
+    });
```
musnit, line 22: oh cool the client already handles using URLs as paths and concurrency and nice progress across urls :O?

src/triggers/ipfs.ts
```diff
@@ -28,3 +28,29 @@ export const unpinnedTrackContent: (cidField: string, limit?: number) => Trigger
 
     return _.uniq(cids);
   };
+
+export const audioNotOnIpfs: Trigger<undefined> = async (clients: Clients) => {
+  const query = `select t.* from "${Table.processedTracks}" as t
+      where "lossyAudioIPFSHash" is null
+      and "lossyAudioURL" is not null
+      LIMIT ${process.env.QUERY_TRIGGER_BATCH_SIZE!}`
```
musnit, line 36: i'm not sure this will work, since the addMetadataIPFSHash processor which runs before this will set the ipfs hashes to '' rather than null.

This is intentional, so that we can distinguish between tracks which haven't had their ipfs hash extracted from their known url vs tracks which don't actually have a hash in their known url.

The reason for this is so that we can preserve a property of the Spindexer that all processors should ideally be commutative, ie, it doesn't matter in which order they're run.

If these 2 processors were to run on a given track before the addMetadataIPFSHash then they would upload some files unnecessarily that are already on IPFS, so we should ensure these triggers runs only after the addMetadataIPFSHash one by triggering on '' rather than on null.

src/processors/default/ipfsMediaUploader.ts
musnit: i think you can use something like:

```
 if ('lossyAudioURL' in processedTrack) {
 ...
} 
```
to check for the existence of a field on an object and then handle an error if that field doesn't exist as expected

src/processors/default/ipfsMediaUploader.ts
musnit: maybe rename newTriggerItems to tracks also to be more readable?

src/processors/default/ipfsMediaUploader.ts
Geimaj: it appears like it would do that but it still seems to upload them sequentially :/ I might have to figure out a way to upload them concurrently with `.add` instead of using `.addAll`

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,51 @@
...
+function processorFunction(sourceField: 'lossyAudioURL' | 'lossyArtworkURL', replaceField: 'lossyAudioIPFSHash' | 'lossyArtworkIPFSHash') {
+  return async (newTriggerItems: any, clients: Clients) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const tracksByFilePath = _.keyBy(newTriggerItems, (track: ProcessedTrack) => new URL(track[sourceField]).pathname.split('/').pop());
+      
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const sources = newTriggerItems.map((track: ProcessedTrack) => {return urlSource(track[sourceField]!)} );
+    const allFiles = await clients.ipfs.client.addAll(sources, {
```
musnit, line 19: hm we await for all here? so if 1 of 100 of the source urls is broken/massive/inaccessible, it will block all the others? i think ideally we should allow for their concurrency to work in parallel such that if a few urls time out, the rest are still processed properly?

musnit: if we change the flow to handle download->upload->update to happen for a single track all in a row instead then you then also won't need this weird hack where you use the filePath to link a completed file download to a track (the filePath hack seems like it could have some edge cases where it breaks too, imi not sure?)

Geimaj: yeah probably another reason to use `.add` instead of `.addAll` and do it manually, and insert any errors into the `nftProcessErrors` table

musnit: To clarify what I mean here is i imagined something like.

```
tracks = [0,1,2,...100]

processTrack = track => { file = await ipfsAddTrack(track); updates.push({ ... file ...)}

promises = await tracks.map(processTrack)

await Promise.all(promises)

db update
```

src/processors/default/ipfsMediaUploader.ts
musnit: that concurrent with .add seems maybe better

src/processors/default/ipfsMediaUploader.ts
Geimaj: so rather than uploading the files using `urlSource` we should download each file and then upload it from our local copy?

musnit: i'm not sure how urlSource works? i assumed it does dl the file and upload it under the hood, but i guess maybe it just tells the ipfs node to actually dl the file? either way, i think using urlSource as you do makes sense/is better as long as it's parallellized as we've discussed. so ignore my comment on downloading the file if it's not relevant.

src/triggers/ipfs.ts
Geimaj: it looks like the addMetadataIPFSHash processor only sets `metadataIPFSHash` to "", and doesn't change the audio or artwork hash. When I query the DB for lossyAudioIPFSHash = null I get the expected results but when querying for "" no results are returned?

musnit: ah yeh you're right - i got confused the the metadata ipfs hash. i think for audio/artwork, their cid is always created at time of processedTrack creation via the platform type if it exists, so this is not an issue.

so your existing code should be fine, nvm sorry

src/processors/default/ipfsMediaUploader.ts
Geimaj: I tried this but I get the error:

Element implicitly has an 'any' type because expression of type '"lossyAudioURL" | "lossyArtworkURL"' can't be used to index type 'ProcessedTrack'.
  Property 'lossyAudioURL' does not exist on type 'ProcessedTrack'


I notice we don't access processedTrack.lossyAudioURL or processedTrack.lossyAudioIPFSHash anywhere else in the codebase so this seems like a problem with how TS is managing those optional fields?

musnit: yeh so basically with TS you cannot access optional fields unless you know that they are there. using object.field! doesn't work because ! just tells TS that this field is not undefined, it does not tell TS that the field actually exists (existence is kinda a separate thing to being undefined)

so for example

a= {}
b = {name:undefined}

these are actually 2 different things, i believe both in JS and TS but definitely in TS so a.name! won't work because TS knows that name is not a field on this object.

the if (field in object) check should allow you to get typescript to confirm whether or not a field actually is there, so for example

```
 if ('name' in a) {
   a.name!
} 
```

this should work afaik, as long as you only access the field within the conditional.

so in our case, did you try wrap all the logic/accesses you want inside a 

```
 if ('lossyAudioURL' in processedTrack) {
 ...
} 
```

block?

(also, i'm not sure about whether `if(sourceField in processedTrack)` works or only a hardcoded `if('lossyAudioURL' in processedTrack)` works, hopefully both?)

and yeh this is one of the weirded edge cases of typescript so if the above suggestion doesn't work then sure let's just leave it as you've already done.

Skin881kd (approved):

src/processors/default/ipfsMediaUploader.ts
Geimaj: yeah unfortunately I didn't have any success with that technique so just leaving the comments in for now

.env.example
```diff
@@ -22,3 +22,4 @@ IPFS_PIN_MAX_PROMISES_PER_MINUTE=60
 POSTGRAPHILE_SERVE_PORT=6001
 NUMBER_OF_ERROR_RETRIES=5
 SOLANA_PROVIDER_ENDPOINT=https://solana-mainnet.g.alchemy.com/v2/key
+IPFS_NODE_MULTIADDRESS=/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr,/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr
```
musnit, line 25: we expect a locally running ipfs node now, right? can you add this to README/docs

musnit, line 25: we have a few different IPFS related env things now. can we add a comment on each to clearly show what they're used for?

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,77 @@
...
+      ['whereIn', ['url', tracks.map(track => track[sourceField])]]
+    ]);
+    const fileByUrl = _.keyBy(existingFiles, 'url')
+    
+    const processTrack = async (track: ProcessedTrack) => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      const url = track[sourceField];
+      try {
+        const source = urlSource(url)
```
musnit, line 30: this is only used within the if? should this line be within the if then?

src/clients/ipfs.ts
```diff
@@ -43,7 +46,8 @@ const init = async () => {
   return {
     getHTTPURL: (ipfsHash: string) => {
       return `${process.env.IPFS_ENDPOINT}${ipfsHash}`;
-    }
+    },
+    client: create()
```
musnit, line 50: can we make this configurable to a url and not only support localhost

src/triggers/ipfs.ts
```diff
@@ -28,3 +28,35 @@ export const unpinnedTrackContent: (cidField: string, limit?: number) => Trigger
...
+      where "lossyAudioIPFSHash" is null
+      and "lossyAudioURL" is not null
+      and i.error is null
+      LIMIT ${process.env.QUERY_TRIGGER_BATCH_SIZE!}`
+
+  const tracks = (await clients.db.rawSQL(
+    query
+  )).rows
+
+  return tracks
```
musnit, line 45: this is with the join, so maybe tracksWithFiles ?

src/triggers/ipfs.ts
```diff
@@ -28,3 +28,35 @@ export const unpinnedTrackContent: (cidField: string, limit?: number) => Trigger
...
+      where "lossyArtworkIPFSHash" is null
+      and "lossyArtworkURL" is not null
+      and i.error is null
+      LIMIT ${process.env.QUERY_TRIGGER_BATCH_SIZE!}`
+
+  const tracks = (await clients.db.rawSQL(
+    query
+  )).rows
+
+  return tracks
```
musnit, line 61: this is with the join, so maybe tracksWithFiles ?

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,77 @@
...
+  return async (tracks: ProcessedTrack[], clients: Clients) => {
+    const updates: ProcessedTrack[] = [];
+    const ipfsFiles: IPFSFile[] = [];
+
+    const existingFiles = await clients.db.getRecords<IPFSFile>(Table.ipfsFiles, [
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      ['whereIn', ['url', tracks.map(track => track[sourceField])]]
+    ]);
+    const fileByUrl = _.keyBy(existingFiles, 'url')
```
musnit, line 23: you're already joining in the trigger with files, so the url/cid mapping should already be available without having to do this extra db query right?

because the processor is actually getting an array of tracksWithFiles, not just tracks ?

src/processors/default/ipfsMediaUploader.ts
```diff
@@ -0,0 +1,77 @@
...
+            ...track,
+            [replaceField]: fileForUrl.cid
+          })
+        }
+      } catch (e: any){
+        ipfsFiles.push({ url, error: e.message });
+      }
+    }
+    
+    await rollPromises<ProcessedTrack, void, void>(tracks, processTrack, 300, 60)
```
musnit, line 58: how do you imagine/expect things like large files, timeouts, etc to get handled in production and operationally?

src/triggers/ipfs.ts
Geimaj: we join on files to only get the tracks without file upload errors, but we only select the columns from tracks so this is still really just tracks

src/processors/default/ipfsMediaUploader.ts
Geimaj: At the moment the processor does just get the track data. I added this extra query to see if any of the files for those tracks have already been uploaded. I originally designed the trigger and the processor to just work with tracks but I'll include the data from the join so we can remove the extra query

src/processors/default/ipfsMediaUploader.ts
Geimaj: There is a timeout variable in the config, if the request timeout is reached then an error will be written to the error column of the ipfsFiles table and that track will be skipped until that error is cleared

musnit: hmm didn't see which timeout variable applies to this?

also, do you know if it handles each of the cases correctly, eg:
 - (A) add a track from a source that is really slow, so it takes very long to add but it is stable and makes progress and so should correctly finish
 - (B) add a track from a source that is fast, but the track is a 40min audiovisual high quality video, so like 1gb, so it takes long but is fast and stable and so should correctly finish
 - (C) add a track from a source that is unavailable or cuts out, so is a genuine error and should timeout and error

I suppose we should have a maximum of a few minutes irrespective of the above scenarios, since we don't want the processor to block on one big track, so even if it's case A or B maybe it should still timeout and we should handle via error?

Geimaj: so far I've only tested with case B and C. I imagine A would cause a timeout too but I will try find a suitable example to test that with. Let me know if you are aware of any

.env.example
```diff
@@ -1,24 +1,42 @@
...
+# IPFS pinning config (Used for pinning existing content on IPFS)
 IPFS_PIN_REQUEST_TIMEOUT=5000
 IPFS_PIN_URL=https://api.pinata.cloud/psa/pins
 IPFS_PIN_JWT=asdf
-MAX_PROMISES_PER_MINUTE=1000
 IPFS_PIN_MAX_PROMISES_PER_MINUTE=60
+
+#IPFS upload config (Used for uploading new content to ipfs)
+IPFS_NODE_MULTIADDRESS=/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr,/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr
+IPFS_NODE_HTTP_API=http://localhost:5001/api/v0
```
musnit, line 36:
```suggestion
IPFS_PIN_MAX_PROMISES_PER_MINUTE=60
IPFS_NODE_MULTIADDRESS=/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr,/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr

#IPFS upload config (Used for uploading new content to ipfs)
IPFS_NODE_HTTP_API=http://localhost:5001/api/v0
```

.env.example
```diff
@@ -1,24 +1,42 @@
...
+
+# IPFS pinning config (Used for pinning existing content on IPFS)
 IPFS_PIN_REQUEST_TIMEOUT=5000
 IPFS_PIN_URL=https://api.pinata.cloud/psa/pins
 IPFS_PIN_JWT=asdf
-MAX_PROMISES_PER_MINUTE=1000
 IPFS_PIN_MAX_PROMISES_PER_MINUTE=60
+
+#IPFS upload config (Used for uploading new content to ipfs)
+IPFS_NODE_MULTIADDRESS=/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr,/ip4/123.456.78.90/tcp/4001/ipfs/QmAbCdEfGhIjKlMnOpQrStUvWxYzAbCdEfGhIjKlMnOpQr
```
musnit, line 35: maybe IPFS_NODE_MULTIADDRESS should go in the pinning section?

README.md
```diff
@@ -19,6 +19,7 @@ Spindexer works by:
  - Postgres running
  - postgresql-client installed
  - ts-node installed globally
+ - ipfs installed
```
musnit, line 22: i guess since we now take in the IPFS_NODE_HTTP_API, having it locally installed is not necessary anymore, so we can remove this?

```suggestion
```

README.md
```diff
@@ -36,6 +37,10 @@ Spindexer works by:
    - IPFS pinning api url for pinning ipfs content to your own node
```
musnit, line 36: maybe just change the comments here to something like:
   - IPFS gateway endpoint for accessing IPFS content
   - IPFS RPC endpoint for adding new content to IPFS
   - IPFS pinning api url for pinning ipfs content to a node

```

```

README.md
```diff
@@ -36,6 +37,10 @@ Spindexer works by:
...
  - Bootstrap the DB with a recent backup so that you don't have to index from scratch:
    - ```yarn restore-db```
+ - Ensure the ipfs node is running:
+   - `ipfs daemon`
+ - Add your ipfs node multiaddress to .env. Find this by copying the address with a public ip after running:
+   - `ipfs id`
```
musnit, line 42: can remove since no longer a local dependency?
```suggestion
```

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/index.ts",
   "scripts": {
     "lint": "npx eslint . --fix",
-    "start": "NODE_EXTRA_CA_CERTS=./db-ssl-certificate.pem ts-node src/index.ts",
+    "start": "NODE_EXTRA_CA_CERTS=./db-ssl-certificate.pem ts-node src/index.ts 2>&1 | tee /tmp/debug",
```
musnit, line 8: we use yarn start in production too, with extra command stuff that logs to ./log and is already hooked into logging, so let's use a different command for this or remove it? we can potentially improve this later

src/types/nftProcessError.ts
```diff
@@ -2,6 +2,7 @@ export type NFTProcessError = {
   nftId: string;
   processError?: string;
   metadataError?: string;
+  ipfsMediaError?: string;
```
musnit, line 5: is this used?

src/processors/default/ipfsMediaUploader.ts
musnit: seems fine for now. if we encounter it in prod we can deal with it then

package.json
Geimaj: moved to start-logs

PR #45, opened 2022-08-17
create consumable views

Geimaj:
add a migration to create a view for each table as `consumable${tableName}`. Each view has the necessary smart comments to enable references between views. 

The only additional constraint at the moment is on the `consumableProcessedTracks` view which requires both `lossyAudioIPFSHash` and `lossyArtworkIPFSHash` to not be null


src/db/migrations/17-add-consumable-views.ts
```diff
@@ -0,0 +1,94 @@
+
+import { Knex } from 'knex';
+
+import { Table } from '../db';
+
+async function getForiegnKeys(knex: Knex): Promise<{
```
musnit, line 6: getForeignKeys

musnit:
Thoughts:
 - Oh kinda nice that it's all automatically generated with an option to add in filters. Any thoughts on the trade off of doing it like this vs just explicitly doing it only for the tables we want to, but maybe autogenerated on all is fine for now?
 - Thinking about API/Naming, especially from the perspective of an outside developer, maybe we should just the base names (ie, Track, Artist) as the view names and migrate the underlying tables to use something like rawTrack, rawArtist so that the naming on the API is simpler?

Geimaj:
> * Oh kinda nice that it's all automatically generated with an option to add in filters. Any thoughts on the trade off of doing it like this vs just explicitly doing it only for the tables we want to, but maybe autogenerated on all is fine for now?

There was a lot of repeated code and room for mistakes when doing it manually, so I decided to take the automatic approach to make sure everything is covered and created correctly

> * Thinking about API/Naming, especially from the perspective of an outside developer, maybe we should just the base names (ie, Track, Artist) as the view names and migrate the underlying tables to use something like rawTrack, rawArtist so that the naming on the API is simpler?

Are we able to make naming changes to the schema and still be backwards compatible with the current API? will /v2 be running on a separate DB?

src/db/migrations/20-add-consumable-views.ts
```diff
@@ -0,0 +1,128 @@
...
+  processors = 'processors',
+  metaFactories = 'metaFactories',
+  nftFactories = 'nftFactories',
+  nfts_processedTracks = 'nfts_processedTracks',
+  nftProcessErrors = 'nftProcessErrors',
+  ipfsPins = 'ipfsPins',
+}
+
+function tableNameToViewName(tableName: string): string {
+  return `consumable${tableName[0].toUpperCase() + tableName.substring(1)}`
```
musnit, line 50: do we still need the consumable prefix for the views now? or can remove?

src/triggers/missing.ts
```diff
@@ -12,7 +12,7 @@ export const ethereumMissingCreatedAtTime: (tableName: Table) => Trigger<undefin
 };
 
 export const missingMetadataObject: Trigger<undefined> = async (clients) => {
-  const nftQuery = `select * from ${Table.nfts} n
+  const nftQuery = `select * from "${Table.nfts}" n
```
musnit, line 15: so theres 2 cases that could result in issue from this PR change.
1 - as in this example, where we need "" around a table name since its got uppercase now
2 - if there are any hardcoded table names anywhere else in the code that don't refer to the Table object

did you run through things pretty comprehensively to ensure we've covered both 1 and 2 in all files? i'll double check over too

also, this makes me think this PR could also a good opportunity for us to rename to snake_case instead of camelCase for SQL coz camelCase seems consistently annoying in sql lol?

src/db/migrations/20-add-consumable-views.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  // specify conditions to add to each view
+  const conditions: {
+    [table in Table]?: string
+  } = {
+    [Table.processedTracks]: `
+    "lossyArtworkIPFSHash" is not null 
+    and "lossyAudioIPFSHash" is not null`,
+  }
```
musnit, line 71: we should also actually ensure that both the audio and artwork are pinned, so join both cids with the pin table and ensure they have an entry there

src/triggers/missing.ts
Geimaj: yeah I checked for all hardcoded table names and replaced with a reference to Table, but also wrapped with "". And yeah I think changing to snake_case is a good idea that will make this less annoying

Geimaj: probably should open another PR renaming field names to snake_case too

src/db/migrations/20-add-consumable-views.ts
```diff
@@ -62,13 +62,22 @@ export const up = async (knex: Knex) => {
...
+    select t.* 
+    from "${Table.processedTracks}" t
+    join "${Table.ipfsPins}" p
+    on t."lossyAudioIPFSHash" = p.id 
+    join "${Table.ipfsPins}" p1
+    on t."lossyArtworkIPFSHash" = p1.id 
+    where "lossyArtworkIPFSHash" is not null 
+    and "lossyAudioIPFSHash" is not null
+    and p.status = 'pinned'
+    and p1.status = 'pinned'
```
musnit, line 79: status isn't reliable (we don't check and update it after pinning) so lets leave that out for now

src/db/migrations/20-add-consumable-views.ts
```diff
@@ -126,6 +129,10 @@ export const down = async (knex: Knex) => {
   for (const table of Object.values(Table)){
     await knex.raw(`drop view "${tableNameToViewName(table)}"`);
   }
+  
+  await knex.schema.alterTable(Table.ipfsPins, table => {
+    table.string('status')
```
musnit, line 134: missing the drop?

Geimaj: the drop is in the up. The down adds the column back

src/db/migration-helpers.ts
```diff
@@ -12,29 +12,14 @@ type NftFactoryAddress = {
 }
 
 export const addPlatform = async (knex: Knex, platform: MusicPlatform) => {
-  await knex.raw(`ALTER TABLE platforms drop constraint "platforms_type_check"`);
+  await knex.raw(`ALTER TABLE "${Table.platforms}" drop constraint "platforms_type_check"`);
```
musnit, line 15: @Geimaj actually is this still platforms_type_check or will it become like raw_platforms_type_check?

musnit: hm platforms_type_check is just the name of the constraint, so should be fine to leave as is i think?

Issue #48, opened 2022-08-22
Add mechanism to verify/approve high quality music nfts from open contracts like Zora's original contract/Rarible/Opensea so we can filter out random audio files and experiments from real music


Geimaj:
Maybe have approved/unapproved on the Metafactory, NFTFactory and NFT records, where approval cascades through all 3.

Eg:
 - Sound.xyz gets MetaFactory approval so that all its generated artist contracts and NFTs are automatically approved
 - NOIZD NFTFactory and most other current manual smart contracts also all get approval
 - Original ZORA contract does not get approval, but during Catalog filter they get approved
 - ZORADrops Metafactory does not get approval, requires manual approval on each NFTFactory

Will also need a way to backup/save approvals in case DB is recreated - maybe some kind of separate seed is needed.
Will also need API/UI for easy manual approval (probably via retool)

When thinking about this mechanism, also imagine it in a scenario where we have a full ERC721NFT indexer that indexes every single contract on Ethereum, irrespective of whether they're music NFTs or not - Ideally the curation/approval mechanism should be capable of supporting this kind of scale too.

PR #49, opened 2022-08-22
Geimaj/approved entities

Geimaj:
implements https://github.com/spinamp/spindexer/issues/48


src/db/migrations/19-add-approved-entities.ts
```diff
@@ -0,0 +1,37 @@
+
```
musnit, line 1: this migration to add the approved field is coming after the migrations to add all the platforms? so the approved field won't really exist yet in the past migrations?

Geimaj: ah yeah, will include it int the bootstrap file too

musnit:
@Geimaj looks like the views don't get updated with the new approved fields once the raw tables have them. need to update views too i guess ... hmm seems like it doesn't actually save the view query with *, but rather with individual fields. Can you look into if there is a more general way to solve this?

Geimaj:
@musnit now when we modify tables in migrations we can call `updateViews` which will update the views and their references

musnit:
> @musnit now when we modify tables in migrations we can call `updateViews` which will update the views and their references

cool! can we add something to the readme to make it clear that all schema migrations should call the updateViews method?

musnit (approved):

Issue #41, opened 2022-08-15
error/retry processor not working properly


musnit:
The retry process doesn't work properly - the retry count never seems to tick up above 1, so the retries happen now every time, indefinitely instead of stopping at 5 retries.

May be related to how processPlatformTracks upserts data - we should double check what happens there and ensure that the counter is not being reset here.

PR #51, opened 2022-08-24
Improve error retries

puffo:
- [x] renames error-related args for improved readability
- [x] fixes  `numberOfRetries` resetting to 1 with upsert

The issue did in fact relate to `processPlatformTracks` overwriting the number of retries with the upsert operation. 

The fix extends our ability to specify which columns to merge when performing an upsert, as per:
https://knexjs.org/guide/query-builder.html#merge

and applies it to `processPlatformTracks`.

Resolves: #41 #50


Geimaj (commented):
Looks good, but can you please check for other usages of upsert? I think `src/processors/default/addMetadataObject.ts` is another example

musnit:
yep looks good, same comment as jamie

and otherwise if you're tested and you can confirm the error ticks up to 5 over 5 runs of the script and then those nfts are skipped thereafter, then should be good  :)!

puffo:
Ah, I missed that, great catch! I updated `addMetadataObject.ts` and the metadata errors are now populated correctly.

I also see that upsert is used in a couple of other places but they're unrelated to NFTProcessErrors. The default merge behaviour will kick in for those upserts i.e. merge everything.

In `createProcessedTracksFromAPI.ts` and `processPlatformTracks.ts`
```ts
  await clients.db.upsert(Table.artists, artists);
  await clients.db.upsert(Table.artistProfiles, (artistProfiles as unknown as Record[]), ['artistId', 'platformId']);
  await clients.db.upsert(Table.processedTracks, mergedProcessedTracks);
```

Another one in `addTimestampFromMetadata.ts` but it looks like it splats the existing NFT field args :)

I think these are working as expected for their use cases, right @musnit @Geimaj ?

musnit (approved):
looks good! we're refactoring a bunch in the other bigger open PRs so may be easier to wait until those are in to merge this, but i good to go otherwise

PR #53, opened 2022-08-25
Geimaj/zora

Geimaj:
(requires rebuild)
index zora nfts

split log queries into batches if getLogs fails
split inserts into batches


src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -66,7 +67,9 @@ const processorFunction = (contracts: NftFactory[]) =>
     const existingNfts = new Set((await clients.db.getRecords<NFT>(Table.nfts, [ ['whereIn', [ 'id', transferNftIds ]] ])).map(nft => nft.id));
     const transfersForExistingNfts = transfers.filter(transfer => existingNfts.has(transfer.nftId!));
 
-    await clients.db.insert(Table.erc721Transfers, transfersForExistingNfts);
+    await clients.db.insert(Table.erc721Transfers, transfersForExistingNfts, {
+      ignoreConflict: ['id', 'from', 'to', 'contractAddress', 'tokenId', 'transactionHash']
```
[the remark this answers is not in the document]
musnit: hmm this is weird - may be worth investigating properly? did you work out why?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -76,7 +79,8 @@ export const createERC721NFTsFromTransfersProcessor: (contracts: NftFactory[]) =
     trigger: newERC721Transfers(
       contracts
         .filter(c => c.standard === NFTStandard.ERC721) //only include ERC721 contracts
-        .map(c => ({ address: c.address, startingBlock: c.startingBlock! })) // map contracts to EthereumContract
+        .map(c => ({ address: c.address, startingBlock: c.startingBlock! })), // map contracts to EthereumContract
+      process.env.ETHEREUM_BLOCK_QUERY_GAP! // todo: overwrite to 2k?
```
musnit, line 83: is todo still relevant?

Geimaj: nope, I removed it. This is handled with the dynamic getLogs query now

src/db/migrations/22-add-zora-creator.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const up = async (knex: Knex) => {
+  await addMetaFactory(knex, ZORA_CREATOR_FACTORY);
+
+  const hasColumn = await knex.schema.hasColumn(Table.erc721Transfers, 'transactionHash')
+  if (!hasColumn){
+    await knex.schema.alterTable(Table.erc721Transfers, table => {
+      table.string('transactionHash');
+      table.dropPrimary(`${Table.erc721Transfers}_pkey`)
+      table.primary(['id', 'from', 'to', 'contractAddress', 'tokenId', 'transactionHash'])
```
musnit, line 27: this change doesn't work due to this issue:
 - column "transactionHash" of relation "raw_erc721_transfers" contains null values

if we want to use transactionHash in the primary key then i guess we should do a full rebuild anyway, so this optional migration is probably unnecessary

musnit: alternatively we could leave it out the primary key, then fill it in, then migrate it into a primary key after

src/clients/ethereum.ts
```diff
@@ -22,9 +22,16 @@ export type EthCall = {
   callInput?: string,
 }
 
+type returnType = ethers.utils.LogDescription & {
```
musnit, line 25: hm what's this used for?

src/clients/ethereum.ts
```diff
@@ -34,8 +41,38 @@ export type ContractFilter = {
...
+      events.push(...eventRange)
+      readUntil = end;
+      start = end;
+      end = BigNumber.from(toBlock).toHexString();
+    } catch (e: any){
+      const errorString = e.toString() as string;
+      const suggestion = errorString.substring(errorString.indexOf('this block range should work: ['), errorString.indexOf(']\\"}}",'));
+      const suggestedRanges = suggestion.substring(suggestion.indexOf('[') + 1).split(', ')
+      start = suggestedRanges[0];
+      end = suggestedRanges[1];
```
musnit, line 66: not ideal if we're implementing a dependency on alchemy here, but it is nice to have this guiding range. i'd say to keep this but we should add more explicit checks on the error string to confirm it is in fact this exact error, and it contains exactly this string so we can throw a proper error if alchemy isn't being used.

similarily, ensuring that if alchemy for example changes there error message or structure of where they put the guiding block range or makes it json or whatever, we make sure we detect this and throw a proper error too

src/db/migrations/22-add-zora-creator.ts
Geimaj: maybe I could also make transactionHash nullable?

src/clients/ethereum.ts
Geimaj: it's the return type for getEventsFrom. I renamed it to Events

src/db/migrations/22-add-zora-creator.ts
musnit: hmm i would expect this primary key to break in the following scenario anyway tho:
 - if someone sends an nft from person a -> b -> a again all in the same transaction, then all 6 of those fields will be the same even though there were 2 different transfer events

and so i would expect a crash again in this scenario

this is why i used the event index as part of the ID to try be super robust/futureproof, so it is weird that we're getting the collision issue. maybe there's a bad assumption i'm making or some bug somewhere. 

did you find out where the pk conflict came from yet?

src/types/metaFactory.ts
```diff
@@ -47,7 +47,15 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
       startingBlock: event.blockNumber,
       contractType: NFTContractTypeName.default,
       standard: NFTStandard.ERC721,
-      autoApprove: approved
+      autoApprove: approved,
+      typeMetadata: {
+        overrides: {
+          artist: {
+            artistId: event.args!.creator,
```
musnit, line 54: ooh this and the mapper looks cool but maybe lets put it in a separate PR so we can get the basic nft indexing merged in first

src/db/migrations/22-add-zora-creator.ts
Geimaj: even in that case wouldn't the log index be different? I'm not sure in exactly which case this was happening but I'll try to find an example

src/types/metaFactory.ts
Geimaj: I think this has to included in this PR, because the type metadata needs to be set at the time of indexing. If we index the factories and then come back later with type overrides, we won't have access the the events anymore and won't be able to set this data unless we reprocess everything?

src/db/migrations/22-add-zora-creator.ts
musnit: yeah the log index will be different, so that's why the log index seems important to use in the pk

Issue #56, opened 2022-08-29
new heds


Geimaj:
https://heds.app/listen/heds/collabtape/goodsociety

looks like its 10 different tracks with 10 different artists, multiple nfts per track.

i think we can do this:
 - dedup nfts using the artist field
 - get artist from artist field 
 - track title doesn't appear to be in the nft, maybe just need to hardcode it

hm may need a custom mapper to match track title with artist :S

PR #57, opened 2022-08-29
add heds collab

Geimaj:
#56


src/types/platforms-types/heds-collab.ts
```diff
@@ -0,0 +1,142 @@
...
+    throw `Can't map artist to track name`;
+  }
+
+  return track
+
+}
+
+function getAvatarFromArtist(artist: string): string {
+  const avatarByArtist: any = {
+    'LOPHIILE FT. TYLER JAY': 'https://firebasestorage.googleapis.com/v0/b/heds-34ac0.appspot.com/o/tapes%2FcollabTAPES%2FGood%20Society%2FImages%2Fartists%2FLophiile.jpg?alt=media&token=538507ad-08d7-499d-a7af-f689377f3349',
```
musnit, line 38: these urls look risky, especially with there being an access token in them. lets upload to ipfs and use those

src/types/platforms-types/heds-collab.ts
```diff
@@ -0,0 +1,142 @@
...
+  contract?: NftFactory,
+  trackId?: string,
+): ProcessedTrack => {
+  if (!contract) {
+    throw new Error(`Contract missing for mapTrack for nft ${nft.id}`)
+  }
+
+  const artist = getTrait(nft, 'Artist');
+
+  const track: Partial<ProcessedTrack> = {
```
musnit, line 88: maybe we can minimize this code by importing a base mapTrack from another type and just extending it with the special fields?

src/types/platforms-types/heds-collab.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (!nft) {
+    throw new Error(`NFT missing for mapArtistProfile for nft`)
+  }
+  if (!contract) {
+    throw new Error(`Contract missing for mapArtistProfile for nft ${nft.id}`)
+  }
+
+  const artist = getTrait(nft, 'Artist');
+
+  return {
```
musnit, line 117: same as above comment

src/types/platforms-types/heds-collab.ts
```diff
@@ -0,0 +1,142 @@
...
+    platformInternalId: artist,
+    platformId: contract.platformId,
+    avatarUrl: getAvatarFromArtist(artist),
+    websiteUrl: nft.metadata.external_url,
+    createdAtTime: nft.createdAtTime,
+    createdAtEthereumBlockNumber: nft.createdAtEthereumBlockNumber,
+  }
+};
+
+const mapNFTtoTrackID = (nft: NFT): string => {
```
musnit, line 129: lets use title for the track id rather than artist

src/db/migrations/21-add-heds-collab.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from 'knex';
+
+import { NftFactory, NFTContractTypeName, NFTStandard } from '../../types/nft';
+import { MusicPlatform, MusicPlatformType } from '../../types/platform';
+import { addNftFactory, addPlatform, removeNftFactory, removePlatform } from '../migration-helpers';
+
+
+const HEDS_COLLAB_PLATFORM: MusicPlatform =
+  { 
+    id: 'heds-collab',
```
musnit, line 10: hmm is there any easy way we can have these nfts still use heds as the platform instead of introducing a new platform? i guess we'd need to put the custom type onto heds too then, applying only on a specific contract address, or something like that?

musnit: or maybe cleaner - we just remove platform from the base NftFactory entirely and insert it later on? may not be super easy tho?

Geimaj: Not too sure about that because the existing heds platform has a type of `single-track-multiprint-contract` which is used by a few different platforms so having stuff specific to one heds contract in there seems like a bad idea. 

Maybe we can add an optional platform-type field to nftFactory where we can override the platform type from the platform. That way this can still belong to the heds platform but we can process the tracks using the heds-collab platform type?

musnit: yeh this sounds good!

src/types/platforms-types/heds-collab.ts
musnit: cool looks good - not really that impactful but interesting to experiment with how we may be able to compose mappers in a more powerful way with higher order functions like this in future

src/db/migrations/21-add-heds-collab.ts
Geimaj: so to use a platform type override you need to specify the id of the platform of the type you want to use. I needed a reference to the platform and not just the platform type, but I think it's safer this way anyway because it's a fk reference and not just a random string. Things might be a bit confusing around platforms / platform types and the naming so let me know if you have any suggestions for improvement @musnit

musnit: hm ye will look through/think about tmrw

musnit: hmm it seems kinda weird to add an extra platform that's just used to hack into the type detection system. maybe better to modify the type mapper system - i'll take a look into it now

musnit: yeah so i think the right way for us to do this would be to modify the processPlatformTracks processor so that it natively supports overriding the type at the nftFactory level via a typeOverride, rather than adding this extra duplicate platform thing

there's some complexities/refactor that will be needed for it, so maybe we should just release this PR under a different platform as you did originally and do that refactor in a later PR?

musnit: @Geimaj i just took a quick stab at it, incomplete tho: https://github.com/spinamp/spindexer/pull/63

musnit:
closing as being replaced by https://github.com/spinamp/spindexer/pull/63

PR #58, opened 2022-08-30
Fix: exclude default timestamp column with upsert

### LINKED ISSUE #12: bug: upsert overwrites createdAtTime
eg: when processPlatformTracks.ts upserts artist/artist profiles, it changes the createdTime for those artists

puffo:
Resolves #12

### DIFF
--- a/src/db/db.ts
+++ b/src/db/db.ts
@@ -2,6 +2,8 @@ import { Knex } from 'knex';
 
 import { Cursor } from '../types/trigger'
 
+export const defaultTimestampColumn = 'createdAtTime';
+
 export enum Table {
   platforms = 'raw_platforms',
   nfts = 'raw_nfts',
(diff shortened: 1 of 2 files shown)


src/db/db.ts
```diff
@@ -2,6 +2,8 @@ import { Knex } from 'knex';
 
 import { Cursor } from '../types/trigger'
 
+export const defaultTimestampColumn = 'createdAtTime';
```
puffo, line 5: Not sure if we want to rather stick with an enum to allow for future extension? i.e. something like
```javascript
export enum DefaultTimestampColumn {
  createdAtTime = 'createdAtTime'
}
```
Thoughts?

musnit: not sure exactly what you mean / what you imagine future possible extension would be?

puffo: I was thinking about `updatedAtTime`, but in the case of an upsert we'd probably want to affect that column. 

If you're good with just a simple `const`, then let's roll on with [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)!

musnit (approved):

PR #60, opened 2022-08-30
Refer to database via env

puffo:


musnit:
@puffo what's the intention behind this fix? afaik, postgres always has a main db named postres so i'm not sure when this would be changed?

puffo:
> @puffo what's the intention behind this fix? afaik, postgres always has a main db named postres so i'm not sure when this would be changed?

Ah, this change helped me with a clean setup locally. Turns out I didn't need that old `postgres` database! 😂 

 I was aware of the `template0` and `template1` databases that are created by default (as the safety nets I believe), but not the `postgres` database. In that case I'll close this one out. 👍

PR #71, opened 2022-09-01
Fix zora artistids

musnit:
@Geimaj fyi


Geimaj (approved):
Ah right, I didn't include the ethereum/ prefix :+1:

PR #76, opened 2022-09-02
Index artifacts album

puffo:
Adds the 33 Artifacts album, and works on the ability to override title and id resolvers

Functionality
- Make `multi-track-multiprint-contract` specify overrides for title and id
- Remove `mapNFTToTrackIds` promise and rather force an immediate return 

Cleanup/Refactor
- Rename `fieldExtractors` (and associates) to `titleExtractors` and `idExtractors`
- Fix slug generation across all platforms
- Create consistency with `mapNFTToTrackIds` type


src/utils/identifiers.ts
```diff
@@ -0,0 +1,3 @@
+import slugify from 'slugify'
+
+export const strictSlugify = (input: string) => slugify(input, { lower: true, strict: true })
```
musnit, line 3: maybe we can just keep the exported function called slugify too?

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,25 @@
...
+export enum CustomTitleExtractor {
   METADATA_NAME = 'metadata.name',
   ATTRIBUTES_TRAIT_SONG_TITLE = 'attributes.trait.songTitle',
-  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track'
+  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track',
+  DEFAULT = ATTRIBUTES_TRAIT_TRACK
 }
 
-export const fieldExtractors: any = {
+export const titleExtractors: any = {
```
musnit, line 12: if this is just for titles, then i think we can use a specific type since it should always be nft => string?

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -83,13 +77,13 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
-  const trackName = getTrait(nft, 'Track');
+const mapNFTtoTrackID = (nft: NFT, extractor: any): string => {
+  const trackName = strictSlugify(extractor(nft));
   return `ethereum/${formatAddress(nft.contractAddress)}/${trackName}`;
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds = async (nfts: NFT[], dbClient?: DBClient | undefined, extractor?: any): Promise<{ [trackId: string]: NFT[] }> => {
+  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft, extractor));
```
musnit, line 86: lets make extractor a specific type and throw error if its undefined

musnit: its also seems kinda weird that the mapNFTsToTrackIds function cares about a titleExtractor - seems like bad seperation of concerns, especially since this logic is very specific to this type rather than more general. i think the function should just take in the nftFactory and handle the rest internally?

src/processors/default/processPlatformTracks/processNFTFactoryTracks.ts
```diff
@@ -83,14 +83,15 @@ const createTracks = async (
...
+  mapNFTsToTrackIds: (nfts: NFT[], dbClient?: DBClient | undefined, titleExtractor?: any, apiTracksByNFT?: any) => Promise<{
     [trackId: string]: NFT[];
   }>,
   nfts: NFT[],
   dbClient: DBClient,
+  titleExtractor: any,
   apiTracksByNFT: any
   ) => {
-  const trackMapping = await mapNFTsToTrackIds(nfts, dbClient, apiTracksByNFT);
+  const trackMapping = await mapNFTsToTrackIds(nfts, dbClient, titleExtractor, apiTracksByNFT);
```
musnit, line 94: you've changed the order of parameters being sent here. there may be (and i think there are) mappers that expect apiTracksByNFT to be the third argument in this function call, so this may break other stuff. I think just put it at the end, or alternatively we can do a bigger refactor to the mapNFTsToTrackIds function that accepts an object as an argument so that it's not order-dependent

src/types/platforms-types/multi-track-multiprint-contract.ts
puffo: I was also pretty uncomfortable with this, it feels a bit messy. 

But the reason for passing the `extractor` instead of the `nftFactory` is that this function works on multiple NFTs, which (as I understand it) might not all map to the same nftFactory. 

If that's the correct assumption, then passing the NFT along with the function that operates on it to get its title is slightly better than locking down all the NFTs to a single nftFactory.

puffo: I take it there isn't an easy way to get the `nftFactory` that created the `NFT` without querying the DB, right @musnit ? That would clean up the function signature.

src/processors/default/processPlatformTracks/processNFTFactoryTracks.ts
puffo: Yeah, we have a lot of different type annotations for `mapNFTsToTrackIds` across the codebase and I had assumed the more specific ones would override any others. But you're right, this isn't a solid choice right now. I've adjusted the order 👍

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -83,13 +77,13 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
   }
 };
 
-const mapNFTtoTrackID = (nft: NFT): string => {
-  const trackName = getTrait(nft, 'Track');
-  return `ethereum/${formatAddress(nft.contractAddress)}/${trackName}`;
+const mapNFTtoTrackID = (nft: NFT, extractor?: TitleExtractor): string => {
+  const trackName = extractor ? extractor(nft) : String(nft.tokenId);
+  return `ethereum/${formatAddress(nft.contractAddress)}/${slugify(trackName)}`;
 };
```
musnit, line 83: i don't think we should fallback to tokenId here - given the intention of this mapper is to support contracts that print multiple nfts (ie, multiple token ids) per track, using a tokenId here is likely to cause duplicate tracks in future.

i'd say we should just throw an error if it's not provided.

let's also work to make this more robust&generic and so call id an IdExtractor rather than assuming it's always driven by the title

```suggestion
const mapNFTtoTrackID = (nft: NFT, idExtractor?: IdExtractor): string => {
  if (!idExtractor) {
    throw new Error('No idExtractor provided');
  }
  const id = idExtractor(nft);
  if(!id) {
    throw new Error('Id not extracted correctly');
  }
   return `ethereum/${formatAddress(nft.contractAddress)}/${id}`;
};
```

(the slugify should go in the idExtractor i think, since it's up to the extractor to do?)

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -83,13 +77,13 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds = async (nfts: NFT[], dbClient?: DBClient | undefined, apiTracksByNFT?: any, extractor?: TitleExtractor): Promise<{ [trackId: string]: NFT[] }> => {
+  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft, extractor));
```
musnit, line 86: as above, lets take in an idExtractor rather than a titleExtractor

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,30 @@
...
+export type ExtractorTypes = {
+  title?: TitleExtractorTypes
+}
+
+export enum TitleExtractorTypes {
   METADATA_NAME = 'metadata.name',
   ATTRIBUTES_TRAIT_SONG_TITLE = 'attributes.trait.songTitle',
-  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track'
+  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track',
+  DEFAULT = ATTRIBUTES_TRAIT_TRACK
```
musnit, line 11: i'm not sure this is a great default - it's only really legacy and used in src/db/migrations/11-add-daniel-allan-glasshouse.ts

actually i think the only time multi-track-multiprint-contract is used other than in this PR is for glasshouse.

having said that, i do think we want to evolve multi-track-multiprint-contract to become the primary/most commonly used mapper since it should cover a superset of most nft contracts.

given that, i'd say lets:
 - avoid doing a default for now, throw an error if there is no extractor specificed
 - update daniel allen retroactively to make it use the ATTRIBUTES_TRAIT_TRACK title extractor

src/types/platforms-types/multi-track-multiprint-contract.ts
musnit: another thing to think about is how the idExtractor related to titleExtractor. I imagine it could make sense for the default idExtractor to be one that runs the titleExtractor and then slugifies the title.

Though this would mean we need the idExtractor() function can take in the titleExtractor override as a paramater, so the idExtractor knows which titleExtractor to use.

I think this ties in with your comment of "If that's the correct assumption, then passing the NFT along with the function that operates on it to get its title is slightly better than locking down all the NFTs to a single nftFactory" - maybe for each nft we should provide all the overrides it has on a per-nft level (including overrides that may be inherited from their parent contract or parent factory) so that both the mapNFTsToTrackIds and mapNFTtoTrackID and all extractor functions all get the override metadata and so are all flexible to use it however they want.

Do you think this makes sense?

src/types/platforms-types/multi-track-multiprint-contract.ts
musnit: > But the reason for passing the `extractor` instead of the `nftFactory` is that this function works on multiple NFTs, which (as I understand it) might not all map to the same nftFactory.

ah yeah that's a good point. i suppose by that same logic they may not even map to the same extractor, so ideally we should provide the extractor per-nft

> If that's the correct assumption, then passing the NFT along with the function that operates on it to get its title is slightly better than locking down all the NFTs to a single nftFactory.
yeah exactly - as you say

src/db/migrations/26-add-jagwar-twin-artifacts.ts
```diff
@@ -0,0 +1,39 @@
...
+const jagwarTwinAddress = '0xe80c7fee3d41e311e0351244231919e04e699e56';
+
+const ARTIFACTS: NftFactory = {
+  address: '0x84F89f7bAcb20970073Be33F22847e58fbe78992',
+  startingBlock: '15316663',
+  platformId: jagwarTwinAddress,
+  contractType: NFTContractTypeName.default,
+  standard: NFTStandard.ERC721,
+  autoApprove: true,
+  approved: true,
```
musnit, line 18: i think we should leave approved: true (for the contract to be approved) but make autoApprove false since we only want to approve specific nfts from this contract we know are songs.

also, i noticed that when indexing this contract, a bunch of errors show up (when i query the db with `select * from raw_nft_process_errors;` ) of `Error mapping track: TypeError: Cannot read properties of undefined (reading 'startsWith')` -> i'm sure some errors are expecting given the contract is messy, but would be good to get better error messages, which i think will be solved with some of the other comments i made on this PR

src/types/platforms-types/multi-track-multiprint-contract.ts
musnit: as discussed, lets also make mapNFTsToTrackIds  here and in the other platform-types and in platform.ts to return the actual array of trackIds rather than a Promise of them

puffo: Latest approach allows for a generic `Extractor` type. `idExtractor` and `titleExtractor` are both `Extractors`

src/db/migrations/26-add-jagwar-twin-artifacts.ts
puffo: Right, those processing errors relate to the non-audio NFTs on the contract. The latest approach eliminates that TypeError when we choose to autoApprove, but those processing errors remain nevertheless. 

Let's chat about whether to lean on upcoming approvals mechanisms or if we should explore per NFT overrrides to approve the audio contracts :)

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,46 @@
...
+}
+
+export enum TitleExtractorTypes {
   METADATA_NAME = 'metadata.name',
   ATTRIBUTES_TRAIT_SONG_TITLE = 'attributes.trait.songTitle',
-  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track'
+  ATTRIBUTES_TRAIT_TRACK = 'attributes.trait.track',
+}
+
+export enum IdExtractorTypes {
```
puffo, line 14: Not much substance to this just yet, but after looking at all the implementations of `mapNFTtoTrackIDs`, there are a few extractable strategies to deposit here - if we're happy with the approach

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,46 @@
...
+  USE_TITLE_EXTRACTOR = 'useTitleExtractor',
 }
 
-export const fieldExtractors: any = {
```
puffo, line 9: Extractor is more generic instead of specific

src/processors/default/processPlatformTracks/processPlatformTracks.ts
```diff
@@ -62,11 +63,12 @@ const processorFunction = (platform: MusicPlatform) => async (nfts: NFT[], clien
     let nftFactoryType: MusicPlatformTypeConfig;
     try {
       nftFactoryType = getNFTFactoryType(nftFactory, platformType);
+      const extractor = idExtractor(nftFactory);
```
puffo, line 66: Chose to always invoke and return undefined when not required so it is backward compatible for any platform

src/db/migrations/26-add-jagwar-twin-artifacts.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const jagwarTwinAddress = '0xe80c7fee3d41e311e0351244231919e04e699e56';
+
+const ARTIFACTS: NftFactory = {
+  id: '0x84F89f7bAcb20970073Be33F22847e58fbe78992',
+  startingBlock: '15316663',
+  platformId: jagwarTwinAddress,
+  contractType: NFTContractTypeName.default,
+  standard: NFTStandard.ERC721,
+  autoApprove: false,
```
puffo, line 17: discuss whether to do approval via NFT overrides or wait for upcoming approvals

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,46 @@
...
+  if (!idExtractorOverride) {
+    return undefined;
+  }
+
+  if (idExtractorOverride === IdExtractorTypes.USE_TITLE_EXTRACTOR) {
+    return titleExtractor(contract);
+  }
+
+  throw new Error('no other options just yet')
+}
```
puffo, line 46: I think we should leave `slugify` out of here so that this is just another Extractor

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -33,18 +31,10 @@ const mapTrack: MapTrack = (
...
-    throw new Error('unknown extractor override provided')
-  }
-
   const track: Partial<ProcessedTrack> = {
-    id: mapNFTtoTrackID(nft),
-    platformInternalId: mapNFTtoTrackID(nft),
-    title: titleExtractor(nft),
+    id: mapNFTtoTrackID(nft, idExtractor(contract)),
+    platformInternalId: mapNFTtoTrackID(nft, idExtractor(contract)),
+    title: titleExtractor(contract)(nft),
```
puffo, line 37: Use the `idExtractor` and `titleExtractors` appropriately

src/processors/default/processPlatformTracks/processPlatformTracks.ts
musnit, line 66: i think it's safe to assume that all nfts within a single nftFactory will use the same idExtractor.

given that, we could just send the nftFactory into getTrackInputs and into the mapNFTsToTrackIds function and let it handle getting the idExtractor

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,46 @@
...
+
+export const titleExtractor = (contract: NftFactory): Extractor => {
+  const titleExtractorOverride = contract.typeMetadata?.overrides?.extractor?.title;
+  if (!titleExtractorOverride) {
+    throw new Error('unknown extractor override provided')
+  }
+  return titleExtractors[titleExtractorOverride];
+}
+
+export const idExtractor = (contract: NftFactory): Extractor | undefined => {
```
musnit, line 35: todo: remove undefined

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -83,13 +73,19 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
-  return `ethereum/${formatAddress(nft.contractAddress)}/${trackName}`;
+const mapNFTtoTrackID = (nft: NFT, extractor?: Extractor): string => {
+  if (!extractor) {
+    throw new Error('No extractor provided');
+  }
+  const id = slugify(extractor(nft));
+  if (!id) {
+    throw new Error('ID not extracted correctly');
+  }
+  return `ethereum/${formatAddress(nft.contractAddress)}/${id}`;
```
musnit, line 84: I think we should aim to make the idExtractor responsible for the whole id, not just the suffix, so let's implement that for this first case.

src/types/processor.ts
```diff
@@ -41,4 +42,5 @@ export type Processor = {
   initialCursor?: Cursor | undefined;
 };
 
-export type MapTrack = (nft: NFT, apiTrack: any, nftFactory?: NftFactory | undefined, trackId?: string) => ProcessedTrack;
+export type MapTrack = (nft: NFT, apiTrack: any, contract?: NftFactory, trackId?: string) => ProcessedTrack;
+export type MapNFTsToTrackIds = (nfts: NFT[], dbClient?: DBClient, apiTracksByNFT?: any, extractor?: Extractor) => { [trackId: string]: NFT[] };
```
musnit, line 46: given we're refactoring these types now, i think it also makes sense to make the input be a object with each item as a field, rather than each item being a function parameter (similar to how MapArtistProfile is today)

src/types/fieldExtractor.ts
```diff
@@ -1,13 +1,46 @@
...
+export enum IdExtractorTypes {
+  USE_TITLE_EXTRACTOR = 'useTitleExtractor',
 }
 
-export const fieldExtractors: any = {
+export type Extractor = (nft: NFT) => string;
+export type TitleExtractorMapping = Record<TitleExtractorTypes, Extractor>
+
+export const titleExtractors: TitleExtractorMapping = {
   'metadata.name': (nft: NFT) => nft.metadata.name,
```
puffo, line 22: type TitleExtractorTypes = keyof titleExtractors

src/types/mapping.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export type NFTtoTrackIdSource = {
+  nfts: NFT[],
+  dbClient?: DBClient,
+  apiTracksByNFT?: any,
+  contract?: NftFactory
+}
+
+export type MapNFTsToTrackIds = (nftToTrackIdSource: NFTtoTrackIdSource) => TrackMapping;
```
musnit, line 18: NFTtoTrackIdSource and nftToTrackIdSource - "source" seems like too specific a name here, kinda confusing i think? maybe just something simple like "input" is better, so maybe NFTtoTrackIdsInput and nftToTrackIdInput?

Also: For the actual variable name i'd just simplify and call it "input" since in the context it's being used, it's clear that this is the input to the map function. (golang has pretty nice conventions for this https://murfitt.net/blog/short-variable-names-in-go)

Also: The dbClient should be removed now that is is no longer a Promise, since it shouldn't be possible to call the db. I think you'll be fine just removing this without breaking anything, since I don't think it's being used anymore?

```suggestion
export type NFTtoTrackIdsInput = {
  nfts: NFT[],
  apiTracksByNFT?: any,
  contract?: NftFactory
}

export type MapNFTsToTrackIds = (input: NFTtoTrackIdsInput) => TrackMapping;
```

src/types/platforms-types/catalog.ts
```diff
@@ -59,8 +53,13 @@ const mapArtistProfile = ({ apiTrack, nft }: { apiTrack: any, nft?: NFT }): Arti
...
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTtoTrackID = (nft: NFT): string => {
+  const [contractAddress, nftId] = nft.id.split('/');
+  return ethereumTrackId(contractAddress, nftId);
+}
+
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 63:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/chaos.ts
```diff
@@ -55,11 +54,11 @@ const getSong = (nft: NFT) => getTrait(nft, 'Song');
...
   const song = getSong(nft);
-  return `ethereum/${formatAddress(nft.contractAddress)}/${song}`;
+  return ethereumTrackId(nft.contractAddress, song);
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 62:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/heds-collab.ts
```diff
@@ -121,11 +117,11 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
   const track = getTrackNameFromArtist(artist);
-  return `ethereum/${formatAddress(nft.contractAddress)}/${track}`;
+  return ethereumTrackId(nft.contractAddress, track);
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 125:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/mintsongs-v2.ts
```diff
@@ -79,16 +77,16 @@ const selectPrimaryNFTForTrackMapper = (nfts: NFT[]) => {
...
 
-  const nftsByMetadataName = _.groupBy(nfts, (nft) => {
+  const nftsByMetadataName = _.groupBy(nftToTrackIdSource.nfts, (nft) => {
     return nft.metadata.name;
   });
 
-  const nftsByTrackId = _.groupBy(nfts, (nft) => {
+  const nftsByTrackId = _.groupBy(nftToTrackIdSource.nfts, (nft) => {
     return mapNFTtoLatestTrackID(nft, nftsByMetadataName[nft.metadata.name])
   });
```
musnit, line 91:
```suggestion
  const nftsByTrackId = _.groupBy(input.nfts, (nft) => {
    return mapNFTtoLatestTrackID(nft, nftsByMetadataName[nft.metadata.name])
  });
```

src/types/platforms-types/mintsongs-v2.ts
```diff
@@ -79,16 +77,16 @@ const selectPrimaryNFTForTrackMapper = (nfts: NFT[]) => {
...
-  if (!dbClient) {
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  if (!nftToTrackIdSource.dbClient) {
     throw new Error('DB Client not provided to mintsongs mapper')
   }
 
-  const nftsByMetadataName = _.groupBy(nfts, (nft) => {
+  const nftsByMetadataName = _.groupBy(nftToTrackIdSource.nfts, (nft) => {
     return nft.metadata.name;
   });
```
musnit, line 87:
```suggestion
  const nftsByMetadataName = _.groupBy(input.nfts, (nft) => {
    return nft.metadata.name;
  });
```

src/types/platforms-types/mintsongs-v2.ts
```diff
@@ -70,7 +68,7 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
 
 const mapNFTtoLatestTrackID = (nft: NFT, dupNFTs: NFT[]): string => {
   const primaryNFT = selectPrimaryNFTForTrackMapper(dupNFTs);
-  return `ethereum/${formatAddress(primaryNFT.contractAddress)}/${primaryNFT.tokenId}`;
+  return ethereumTrackId(primaryNFT.contractAddress, String(primaryNFT.tokenId));
```
musnit, line 71: minor comment/point to be aware of:

Since tokenId is a bigint, the more canonical way to convert it would be using toString()

It probably won't make a difference in our case, but for reference: https://stackoverflow.com/questions/3945202/whats-the-difference-between-stringvalue-vs-value-tostring

```suggestion
  return ethereumTrackId(primaryNFT.contractAddress, primaryNFT.tokenId.toString());
```

src/types/platforms-types/nina.ts
```diff
@@ -55,16 +53,17 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
     platformId: nft.platformId,
     avatarUrl: undefined,
     websiteUrl: `${nft.metadata.external_url}/related`,
-    createdAtTime: nft.createdAtTime, 
+    createdAtTime: nft.createdAtTime,
   }
 };
 
 const mapNFTtoTrackID = (nft: NFT): string => {
+  // TODO: extract solanaTrackId function and provide appropriate trackId
```
musnit, line 61: we don't really use TODOs in our codebase, and probably shouldn't start

also this is minor enough that i wouldn't create a github issue or anything to track it, so just leave as is i think

```suggestion
```

src/types/platforms-types/nina.ts
```diff
@@ -55,16 +53,17 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
   return `solana/${nft.id}`
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 67:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/noizd.ts
```diff
@@ -52,11 +52,11 @@ const mapNFTtoTrackID = (nft: NFT): string => {
...
   const trackId = externalURL.split('/assets/')[1];
-  return `ethereum/${formatAddress(contractAddress)}/${trackId}`;
+  return ethereumTrackId(contractAddress, trackId);
 }
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 60:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/sound.ts
```diff
@@ -62,11 +60,11 @@ const mapArtistProfile = ({ apiTrack, nft }: { apiTrack: any, nft?: NFT }): Arti
   }
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[], dbClient?: DBClient, apiTracksByNFT?: any): Promise<{ [trackId: string]: NFT[] }> => {
-  if (!apiTracksByNFT) {
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  if (!nftToTrackIdSource.apiTracksByNFT) {
```
musnit, line 64:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  if (!input.apiTracksByNFT) {
```

src/types/platforms-types/sound.ts
```diff
@@ -62,11 +60,11 @@ const mapArtistProfile = ({ apiTrack, nft }: { apiTrack: any, nft?: NFT }): Arti
...
     throw new Error('Expecting apiTracksByNFT for sound mapper');
   }
-  return _.groupBy(nfts, nft => apiTracksByNFT[nft.id]);
+  return _.groupBy(nftToTrackIdSource.nfts, nft => nftToTrackIdSource.apiTracksByNFT[nft.id]);
```
musnit, line 67:
```suggestion
  return _.groupBy(input.nfts, nft => input.apiTracksByNFT[nft.id]);
```

src/types/platforms-types/single-track-multiprint-contract.ts
```diff
@@ -75,11 +73,12 @@ export const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, n
...
-  return `ethereum/${formatAddress(nft.contractAddress)}`;
+  // TODO: rather use ethereumTrackId and provide appropriate trackId
+  return ethereumArtistId(nft.contractAddress);
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
 }
```
musnit, line 82:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
}
```

src/types/platforms-types/single-track-multiprint-contract.ts
```diff
@@ -75,11 +73,12 @@ export const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, n
...
 const mapNFTtoTrackID = (nft: NFT): string => {
-  return `ethereum/${formatAddress(nft.contractAddress)}`;
+  // TODO: rather use ethereumTrackId and provide appropriate trackId
+  return ethereumArtistId(nft.contractAddress);
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft));
```
musnit, line 81:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft));
```

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -83,13 +72,18 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
...
 };
 
-const mapNFTsToTrackIds = async (nfts: NFT[]): Promise<{ [trackId: string]: NFT[] }> => {
-  return _.groupBy(nfts, nft => mapNFTtoTrackID(nft));
+const mapNFTsToTrackIds: MapNFTsToTrackIds = (nftToTrackIdSource) => {
+  if (!nftToTrackIdSource.contract) {
+    throw new Error('No contract provided');
+  }
+  return _.groupBy(nftToTrackIdSource.nfts, nft => mapNFTtoTrackID(nft, nftToTrackIdSource.contract));
 }
```
musnit, line 87:
```suggestion
const mapNFTsToTrackIds: MapNFTsToTrackIds = (input) => {
  if (!input.contract) {
    throw new Error('No contract provided');
  }
  return _.groupBy(input.nfts, nft => mapNFTtoTrackID(nft, input.contract));
}
```

src/processors/default/processPlatformTracks/processNFTFactoryTracks.ts
```diff
@@ -83,16 +84,15 @@ const createTracks = async (
...
 
 export const getTrackInputs = async (
-  mapNFTsToTrackIds: (nfts: NFT[], dbClient?: DBClient | undefined, apiTracksByNFT?: any) => Promise<{
-    [trackId: string]: NFT[];
-  }>,
-  nfts: NFT[],
-  dbClient: DBClient,
-  apiTracksByNFT: any
+  mapNFTsToTrackIds: MapNFTsToTrackIds,
+  nftToTrackIdSource: NFTtoTrackIdSource,
```
musnit, line 88:
```suggestion
  nftToTrackIdInput: NFTtoTrackIdInput,
```

src/processors/default/processPlatformTracks/processPlatformTracks.ts
```diff
@@ -59,14 +60,21 @@ const processorFunction = (platform: MusicPlatform) => async (nfts: NFT[], clien
...
+
     let nftFactoryType: MusicPlatformTypeConfig;
     try {
       nftFactoryType = getNFTFactoryType(nftFactory, platformType);
       const {
         newTrackIds,
         trackMapping,
         existingTrackIds,
-      } = await getTrackInputs(nftFactoryType.mappers.mapNFTsToTrackIds, factoryNFTs, clients.db, apiTracksByNFT);
+      } = await getTrackInputs(nftFactoryType.mappers.mapNFTsToTrackIds, nftToTrackIdSource);
```
musnit, line 77: put dbClient back into getTrackInputs parameters, since it is still needed in getTrackInputs even though not within the mapper

```suggestion
      } = await getTrackInputs(nftFactoryType.mappers.mapNFTsToTrackIds, nftToTrackIdSource, clients.db);
```

src/processors/default/processPlatformTracks/processNFTFactoryTracks.ts
```diff
@@ -83,16 +84,15 @@ const createTracks = async (
...
   ) => {
-  const trackMapping = await mapNFTsToTrackIds(nfts, dbClient, apiTracksByNFT);
+  if (!nftToTrackIdSource.dbClient) {
```
musnit, line 90:
```suggestion
  if (!dbClient) {
```

src/processors/default/processPlatformTracks/processNFTFactoryTracks.ts
```diff
@@ -83,16 +84,15 @@ const createTracks = async (
...
+    throw new Error('No db client provided');
+  }
+  const trackMapping = mapNFTsToTrackIds(nftToTrackIdSource);
   const trackIds = Object.keys(trackMapping);
-  const existingTrackIds = await dbClient.recordsExist(Table.processedTracks, trackIds);
+  const existingTrackIds = await nftToTrackIdSource.dbClient.recordsExist(Table.processedTracks, trackIds);
```
musnit, line 95:
```suggestion
  const existingTrackIds = await dbClient.recordsExist(Table.processedTracks, trackIds);
```

src/db/migrations/26-add-jagwar-twin-artifacts.ts
```diff
@@ -0,0 +1,40 @@
...
+  approved: true,
+  typeMetadata: {
+    overrides: {
+      artist: {
+        name: 'Jagwar Twin',
+        artistId: ethereumArtistId(jagwarTwinAddress),
```
musnit, line 23: i think we should duplicate all artist fields here, since there is something wrong with the upsert of artists that i think is overwriting undefined fields. i think jamie's upcoming PR may fix that but lets just duplicate them here to be safe too

```suggestion
        artistId: ethereumArtistId(jagwarTwinAddress),
                avatarUrl: 'https://web3-music-pipeline.mypinata.cloud/ipfs/QmcBb9C69vvJXasxTYFPwpo9WuZv415KkH3wTdoeBMH2hH',
        websiteUrl: 'https://jagwartwin.com/'
```

src/types/mapping.ts
puffo: Great catch for nuking the db in this object. And I like the naming change, much cleaner.

Awesome on all fronts! 💯

PR #78, opened 2022-09-03
Geimaj/crdt approvals

Geimaj:
create generic mutations interface and experiment with a CRDT. 

Notes:
- I think we should change every entity to have an id field for consistency
- I think the ORM should not change keys and only change values
- I specified the node version in package.json which could break prod unless it has the right version (this is to support the new `node:*` imports)


src/db/migrations/25-zora-approvals.ts
```diff
@@ -0,0 +1,28 @@
+import { Knex } from 'knex';
+
+import { getCrdtUpdateMessages } from '../../types/message';
+import { NftFactory } from '../../types/nft';
+import { Table } from '../db';
+
+export const up = async (knex: Knex) => {
+
+  // approve zora factories
+  const messages = getCrdtUpdateMessages<NftFactory>(Table.nftFactories, {
```
Geimaj, line 10: just for testing / used as an example

src/db/sql-db.ts
```diff
@@ -138,12 +138,16 @@ const init = async (): Promise<DBClient> => {
...
           // exclude the default timestamp column unless explicitly specified in mergeOptions
-          if (mergeOptions === undefined) {
-            mergeOptions = Object.keys(dbUpsert).filter((value) => { return value !== defaultTimestampColumn });
-          }
+          // if (mergeOptions === undefined) {
+          //   mergeOptions = Object.keys(dbUpsert).filter((value) => { return value !== defaultTimestampColumn });
+          // }
+
+          // remove undefined properties
+          const properValues = _.omitBy(dbUpsert, _.isUndefined)
```
Geimaj, line 146: The upsert was overriding data in the table, since the ORM was accessing undefined properties of an object. This fixes that by removing all undefined values. If we want to erase a value, we can explicitly set that value to null in the object passed to `upsert`.

I think this also removes the need for the commented out code above ? @puffo

src/processors/default/messagesProcessor.ts
```diff
@@ -0,0 +1,39 @@
...
+      await clients.db.upsert(message.table, [{
+        id: message.entityId,
+        [message.column]: message.value
+      }])
+      processedMessages.push({
+        messageId: message.timestamp,
+        table: message.table,
+        column: message.column,
+        entityId: message.entityId,
+        processedAt: hrtime.bigint().toString(),
```
Geimaj, line 31: currently using nano seconds to record the processed at time. Probably better to go with a normal timestamp here. Not sure if this field is even necessary at all but it seems nice to have

src/processors/default/messagesProcessor.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  name: NAME,
+  trigger: newMessages,
+  // recieves messages in order of timestamp
+  // messages older than what has already been processed should not be included
+  processorFunction: async (crdtMessages: CrdtMessage[] , clients: Clients) => {
+    const processedMessages: ProcessedMessage[] = []
+
+    // process messages
+    for (const message of crdtMessages){
```
Geimaj, line 21: Need to optimise to perform all updates for a single table at once, rather than one at a time

src/triggers/messages.ts
```diff
@@ -0,0 +1,51 @@
...
+import _ from 'lodash';
+
+import { Table } from '../db/db';
+import { CrdtMessage } from '../types/message';
+import { Trigger } from '../types/trigger';
+
+
+export const newMessages: Trigger<undefined> = async (clients) => {
+  // get all unprocessed messages
+  const selectSql = `select m.* from ${Table.crdtMessages} m
```
Geimaj, line 11: This will only fetch `QUERY_TRIGGER_BATCH_SIZE` rows, so if there are more than `QUERY_TRIGGER_BATCH_SIZE` stale messages the trigger won't actually process properly. Currently as we are the only one creating messages, all messages should be written to processed_messages so this shouldn't be a problem immediately, but will be in the future once we start syncing changes from other nodes. Maybe this trigger should also pass a list of stale messages to the processor, so that the processor can skip processing them and just write them directly to processed messages so that the trigger can ignore them

musnit:
cool will review - maybe we should try do the address -> id change in its own PR to keep it simpler to review and merge?

crdt.md
```diff
@@ -0,0 +1,25 @@
...
+
+Note: this is not a fully p2p design as it relies on the central storage of the messages list.
+
+
+# Messages
+Messages are the mechanism for mutating data within spinamp. Messages are a G-Set, meaning that messages can only be appended to the messages list.
+
+Messages consist of a timestamp to provide order, the dataset, the field, and the value to mutate. Messages are idempotent, meaning that the same messages can be applied multiple times and result in the same mutations being applied and arriving at the same final state.
+Messages are commutative, in that messages can be applied in any order, but will only be applied if a newer timestamp for the same column and id has not been seen yet. This means messages must be processed using a last write wins (LWW) strategy
```
musnit, line 12: Cool, I think this should continue to work even as we decentralize, because in most cases each (id, table) pair can still have a centralized source of messages. Eg: Artist update messages will come from the artist and always be signed by their single key.

Of course it has problems (eg: an artist could accidentally create a far-in-the-future timestamp and fuck up their record permanently), but we'll improve the design to handle that when it becomes more decentralized.

musnit: i do think we should flesh out this doc into a more thorough spec though too.

musnit:
> There is still one open issue which could require bigger changes: If a message is created referencing an ID that hasn't been created yet, it will create the row with the id and value specified because it uses upsert. This means that later if some processor tries to create the same id it could fail, or overwrite some data. This could require us to use this message system for all inserts / updates, and replace the direct writes to the tables.
> 
> We could also potentially separate out insert and update with an extra `operation` field in the message, and make updates via messages fail if the referenced id doesn't exist. Then the message can be applied once the referenced id exists and we don't have to worry about data dependencies or refactor everything to use messages.

yeah i think either of these approaches could work. lets think about/discuss tradeoffs on them on call

src/triggers/messages.ts
musnit: yeah, i think stale messages should still be processed (and just do nothing as part of their processing)

musnit: the existing newMessages trigger i suppose is something like a "restore-from-backup" solution, where the intention is for the table to be append-only and for every message to be processed.

i imagine when we have other nodes and peer with them, we'll have a separate table and trigger for those, like a mempool, where stale messages are just dropped/deleted from the table

src/types/message.ts
```diff
@@ -0,0 +1,36 @@
...
+  processedAt: string;
+}
+
+type Values<T> = {
+  id: string;
+} & Partial<T>
+
+export function getCrdtUpdateMessages<T>(table: Table, values: Values<T>): CrdtMessage[]{
+  const messages: CrdtMessage[] = Object.keys(values).filter(key => key !== 'id').map(key => ({
+    timestamp: hrtime.bigint().toString(),
```
musnit, line 28: yeh lets just use normal timestamps, the same as we do in other parts of the codebase

Geimaj: normal timestamps didn't have enough resolution to be unique and used as an ID in this case, so I opted for nanoseconda as an easy solution.  In the future I think this should have a logical time component and a physical time component so we should use something like an HLC, but this was a quick way to get it working for now so I could begin testing some things

src/db/migrations/24-add-crdt-messages.ts
```diff
@@ -0,0 +1,33 @@
...
+    table.string('value');
+  })
+
+  await knex.schema.createTable(Table.processedMessages, table => {
+    table.string('messageId').primary();
+    table.enu('table', Object.values(Table));
+    table.string('column');
+    table.string('entityId');
+    table.string('processedAt');
+  })
```
musnit, line 21: Idea:

We make this table table/column/entityId centric, ie record-centric and it is used to track the latest update to a given column on a given record, rather than to track anything about the messages that caused the update

musnit:
Summary of review discussion:
```
Stateful Spindexer DB:
 - Contains current active state of the spinamp DB (eg: all current tables)

CRDT processor:
 - Takes new crdt messages from whatever source, processes them and updates the spinamp DB and then can discard those messages as they've been processed and are no longer needed 

(future) Mempool/p2p network of CRDT messages:
 - different nodes or users (eg: an artist) that create messages and share them across the network.
 - an individual node like our own will store these messages temporarily until they're processed by the main crdt processor and then discarded
 - One aspect of the mempool is it may receive messages that are not yet ready (eg: a message with a timestamp in the future, or a message to update a field on a record that does not yet exist). There's a question around what to do with these not-yet-ready messages. They could be discarded or kept but this is for future consideration.

Backup Spindexer Messages: 
 - This is a list of messages that are expected to always be included and always be processed to become part of the Stateful Spindexer DB
 - We store this list in a registry (atm it will be just the crdtMessages table in the db) - maybe should rename this to something more appropriate like backupMessages
 - There are some backup spindexer messages that have the property that they may be for future records that have not yet been indexed, and so for example, may need to wait until some other processor finished indexing those other records before the backup spindexer message can be processed. Given we have the extra requirement on these messages that we want to ensure that they are all processed, we need to make sure that we don't discard, but we also need to make sure they they don't block the main CRDT processor.
   - 2 ideas for how to do this:
     - 1: We add a processedMessages table that tracks what has been processed. We keep the backup messages and don't discard them. We ensure the trigger joins across this table so that unprocessedMessages are

So we have 2 processors:
 - 1: insertSeedsIntoMempool -> has a cursor based on auto-increasing id of backup records. it will insert records from the backup into the mempool.
 - 2: processMempool -> cursorless, it will process records and discard them once processed.

And 3 tables:
 - seeds: has an auto-append backup of seed messages
 - mempool: keeps pending messages still to be processed
 - crdtState: keeps track of latest update across all columns on all records
 - ```

File: src/db/sql-db.ts (hunk 2 above)
Comment (puffo) [reply src/db/sql-db.ts:146]: Yep, I think that's cool, much cleaner too 👍

File: src/db/migrations/26-add-crdt-messages.ts [hunk 9]
```diff
@@ -0,0 +1,47 @@
+import { Knex } from 'knex';
+
+import { CrdtOpetation } from '../../types/message';
```
musnit, line 3: CrdtOperation

src/triggers/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+      and ((rm.operation = '${CrdtOpetation.UPDATE}' and t.id is not null) or rm.operation = '${CrdtOpetation.INSERT}')
+      order by rm."table", rm."column", rm."entityId", rm.timestamp
+      `;
+
+      const result = await clients.db.rawSQL(sql);
+      return result.rows;
+    }
+  }
+
+export const newMessages: Trigger<Cursor> = async (clients, cursor: string) => {
```
musnit, line 28: maybe call this newSeedMessages?

src/db/migrations/26-add-crdt-messages.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Table } from '../db';
+import { tableNameToViewName, updateViews } from '../migration-helpers';
+
+export const up = async (knex: Knex) => {
+  await knex.schema.createTable(Table.seeds, table => {
```
musnit, line 8: should we add an explicit auto-incrementing id field here to use as a primary key?

src/triggers/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+  const selectSql = `select m.* from ${Table.seeds} m
+  limit ${parseInt(process.env.QUERY_TRIGGER_BATCH_SIZE!)}  
+  offset ${parseInt(cursor)}
```
musnit, line 31: maybe we should have an explicity autoincrementing private key and sort by it explicitly instead of just depending on offset?

src/triggers/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+      left outer join ${table} t 
+      on rm."entityId" = t.id
+      left outer join ${Table.crdtState} rcs
+      on rm."table" = rcs."table" 
+      and rm."column" = rcs."column" 
+      and rm."entityId" = rcs."entityId"
+      where rm."table" = '${table}'
+      and ((rm.operation = '${CrdtOpetation.UPDATE}' and t.id is not null) or rm.operation = '${CrdtOpetation.INSERT}')
+      order by rm."table", rm."column", rm."entityId", rm.timestamp
+      `;
```
musnit, line 21: should we have a limit too?

src/triggers/messages.ts
```diff
@@ -0,0 +1,42 @@
+
+import { Table } from '../db/db';
+import { CrdtMessage, CrdtOpetation } from '../types/message';
+import { Cursor, Trigger } from '../types/trigger';
+
+export const pendingMempoolMessages: (tables: string) => Trigger<undefined> = 
+  (table) => {
+    return async (clients) => {
+      const sql = `
```
musnit, line 9: looks good. given this is fairly complex/hard to read, probably should put a comment here describing the intention behind this query and explaining the expected functionality more clearly

src/triggers/messages.ts
```diff
@@ -0,0 +1,42 @@
...
+      select rm.*, rcs."lastTimestamp"
+      from raw_mempool rm 
+      left outer join ${table} t 
+      on rm."entityId" = t.id
+      left outer join ${Table.crdtState} rcs
+      on rm."table" = rcs."table" 
+      and rm."column" = rcs."column" 
+      and rm."entityId" = rcs."entityId"
+      where rm."table" = '${table}'
+      and ((rm.operation = '${CrdtOpetation.UPDATE}' and t.id is not null) or rm.operation = '${CrdtOpetation.INSERT}')
```
musnit, line 19: maybe we should just do inserts and updates in separate triggers/processors to simplify things?

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    name: 'processMempool',
+    trigger: pendingMempoolMessages(table),
+    processorFunction: async (messages: PendingMempoolMessage[] , clients: Clients) => {
+      // ignore inserts for now
+      const staleMessages: MempoolMessage[] = messages.filter(message => message.operation === CrdtOpetation.INSERT);
+
+      // messages are ordered by timestamp.
+      //  group by table, column, entity so that we can easily categorize fresh and stale messages
+      const groupedMessagesUpdates = _.groupBy(
```
musnit, line 21: i'm not sure _.groupBy guarantees preservation of order? so the above assumption may be broken by it

musnit: maybe we should sort by timestamp again just in case

musnit: actually nvm, looks like it doesn't guarantee order of groups, but it does guarantee order of items within a group, so we should be fine

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+  (table) => ({
+
+    name: 'processMempool',
+    trigger: pendingMempoolMessages(table),
+    processorFunction: async (messages: PendingMempoolMessage[] , clients: Clients) => {
+      // ignore inserts for now
+      const staleMessages: MempoolMessage[] = messages.filter(message => message.operation === CrdtOpetation.INSERT);
+
+      // messages are ordered by timestamp.
+      //  group by table, column, entity so that we can easily categorize fresh and stale messages
```
musnit, line 20:
```suggestion
      //  group by table, column, entity so that we can easily categorize fresh and stale messages
      // messages within each group are ordered by timestamp in the trigger, so it's safe to assume the first message of each group is the freshest
```

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+      const groupedMessagesUpdates = _.groupBy(
+        messages.filter(message => message.operation === CrdtOpetation.UPDATE),
+        message => `${message.operation}.${message.table}.${message.column}.${message.entityId}`
+      );
+      const entityUpdates: { [id: string]: { [column: string]: string } } = {};
```
musnit, line 25: hmm this is primarily working for string types? maybe will need to adjust types/have a slightly more complex type system to work for columns of other types too?

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+
+          crdtUpdates.push({
+            table: freshestUpdate.table,
+            column: freshestUpdate.column,
+            entityId: freshestUpdate.entityId,
+            lastTimestamp: freshestUpdate.timestamp,
+          })
+        } else {
+          staleMessages.push(freshestUpdate);
+        }
```
musnit, line 49: i guess we could simplify this nested conditional by:
 - instead of popping for the freshestUpdate, just use groupMessage[0]
 - remove this else condition, since the staleMessages.push will then handle pushing in the first message

src/processors/default/processMempool.ts
musnit: i guess could be complex since we are storing the update as a string in the db already, so we do kinda need to serialize all updates into strings for this table and then deserialize them into the right type for the objectUpdates. still todo to think about best way to handle this

src/db/migrations/26-add-crdt-messages.ts
```diff
@@ -0,0 +1,47 @@
...
+    table.dateTime('timestamp');
+    table.enu('table', Object.values(Table));
```
musnit, line 10: q - should we use an enum for this or just a string?
i guess enum is good in that it gives us a kind of type safety, but means we need to remember to update this enum on all 3 tables whenever we add a new table to the system, similar to how we now need to call updateViews() every migration? i guess that's fine still though - i'm assuming there's no way for the enum to be a type that automatically/dynamically tracks table names?

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+      const groupedMessagesUpdates = _.groupBy(
+        messages.filter(message => message.operation === CrdtOpetation.UPDATE),
+        message => `${message.operation}.${message.table}.${message.column}.${message.entityId}`
+      );
+      const entityUpdates: { [id: string]: { [column: string]: string } } = {};
+      const crdtUpdates: CrdtState[] = [];
+
+      Object.values(groupedMessagesUpdates).forEach(groupMessage => {
+        // gets the freshest message out of the mempool
+        const freshestUpdate = groupMessage.pop()!;
```
musnit, line 30: It would be nice if we could remove the assumption that timestamps don't clash and that clients have synchronised clocks. This would also simplify your zora-approvals migration so that you can remove all the delay stuff. One easy way we could do this right here is to implement a really simple hybrid logical clock that just does the following:
 - filter updates to find all the updates that have the freshest timestamp
 - if theres only 1, great, use it and continue
 - if there are more than 1, sort them by a string sort on the value and use the first one

src/processors/default/processMempool.ts
musnit: could potentially just make this CRDT string-focused only and so not support other types. though if we go that route, would be nice to have some validation/constraint on messages that they only allow columns which reference string types.

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Clients, Processor } from '../../types/processor';
+
+export const processMempool: (table: Table) => Processor = 
+  (table) => ({
+
+    name: 'processMempool',
+    trigger: pendingMempoolMessages(table),
+    processorFunction: async (messages: PendingMempoolMessage[] , clients: Clients) => {
+      // ignore inserts for now
+      const staleMessages: MempoolMessage[] = messages.filter(message => message.operation === CrdtOpetation.INSERT);
```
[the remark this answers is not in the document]
Geimaj: yes still todo, just wanted a review on the new design before finishing all the details

src/triggers/messages.ts
Geimaj: yeah I was also considering that approach. It could be nicer so I'll give it a go

src/processors/default/processMempool.ts
Geimaj: This is built with the assumption that all clients have synchronised clocks. The delays stuff in the zora-approvals migration is just so I can test messages with new and old timestamps, but that won't be necessary in production.

I understand that using a string sort on the value will get us a way to resolve conflicts into a single update, but how can we be sure that that result is the one we want (the one that happened last)? Wouldn't we need something separate to the value to determine the latest message? E.g a lamport clock

src/processors/default/processMempool.ts
Geimaj: The freshest message is at the last position in the array, not at position 0. I think your suggestion could still work though.

I can conditionally push the message into the updates arrays, and then remove the else and add everything to staleMesages. It doesn't matter if a fresh update is pushed into staleMessages because all the messages need to be deleted anyway - just as long as we process it before we delete it

src/db/migrations/26-add-crdt-messages.ts
Geimaj: I think the way to do it dynamically would be to create a table and then use FK references here instead of an enum. That's nice because of the type safety, and we also have a dynamic list of tables that can be used in messages. 

The same table could be referenced here https://github.com/spinamp/spindexer/blob/7c9e3d44e1bc7c8e15ea590bf8af1edade1a6881/src/index.ts#L42-L49

src/processors/default/processMempool.ts
musnit: Well what I'm suggesting is that we're making the choice to say that actually, the first message in the string sort IS the last message and IS the one we want.

So in the case of a clashing timestamp, we're saying we always do want the first value in the string sort.

The only risk/edge I think is if there is some potential edge case scenario where we incidentally/accidentally create 2 updates on the same column with the same timestamp and actually do expect that there is a specific order for them to be processed in? Is there any scenario in which you imagine this happening?

src/db/migrations/26-add-crdt-messages.ts
musnit: hmm ye though we'd still need to make sure that new tables are added to that list of tables lol. i suppose that can be automated but yeh does seem like overkill for now? maybe fine to leave as enum as is?

src/processors/default/processMempool.ts
Geimaj: It should work with any field that can be serialized into a string. I've tested it with string, bigint, dates, json and booleans which should cover all the types used in our db

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    name: 'processMempoolInserts',
+    trigger: pendingMempoolInsertMessages(table),
+    processorFunction: async (messages: PendingMempoolMessage[] , clients: Clients) => {
+      const { crdtUpdates, results } = categorizeMessages(messages);
+
+      const insertResults: { id: string, insert: string }[] = results;
+      const rows = Object.values(insertResults).map(result => JSON.parse(result.insert));
+
+      await clients.db.insert(table, rows)
```
Geimaj, line 82: should probably use upsert here in case a message for the same id is processed in the future?

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,87 @@
...
+      await clients.db.insert(Table.crdtState, crdtUpdates)
```
Geimaj, line 83: probably should also use upsert here

src/processors/default/processMempool.ts
musnit: yes but we need to ensure that every field is overwritten as part of the upsert, even those that are undefined

src/processors/default/processMempool.ts
musnit: ye normal upsert is fine here

src/types/message.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return messages
+}
+
+export function getCrdtInsertMessages<T>(table: Table, id: string, data: T ): CrdtMessage{
+  return {
+    timestamp: new Date(),
+    entityId: id,
+    table,
+    column: 'insert',
```
musnit, line 57: maybe we use 2 separate tables for updates vs inserts since these are quite different types of crdts/functionality, even though they're being used for similar purposes

musnit:
Let's create a seperate issue to deal with in a seperate PR, that will be necessary for opening up the messages API. We should also track this issue as part of the task for opening up the API.

Specifically, some considerations:

error handling todos:
 - invalid/failing message
    - depends on the reason for failing
        - if error message is due to invalidity (eg: upsert for a column that does not exist), we should delete it
        - consideration: we process messages in batched DB calls, so if there is just 1 invalid message in a batch, ideally the other messages should succeed and we should get a clear error pointing to the exact message(s) that failed so that we can then delete those messages. it's unclear if this exact functionality can be achieved through plain existing db queries (eg: some kind of conflict/error handling within the postgres query, and some kind of comprehensive, high fidelity error response). given this, we should investigate the scenario and improve our query to provide this, or if not possible, find an alternative.
        - for now, since all messages are in our control, we'll detect this quickly if a mistake is made and be able to remedy it before deploying to production, but as soon as we open up our messages api this could become a problem
        - an alternative would be to implement robust message validation before or as part of processing them so that we're sure errors are not possible

far-in-the-future-messages:
 - we may want to just ignore these and leave them in the mempool until they're ready, so, the trigger filters out all messages from the future, potentially even deleting them from the mempool.

 - deletions (for future PR/issue)
   - probably need new kind of CRDT, similar to inserts
    - that crdt will have a message and processor that creates a seperate tombstone, in something like a tombstone table for each the record
    - that processor will also then deletes that record
    - update+insert crdt logic will need to be updated to check for tombstones, and discard messages for tombstoned records. For the update trigger, we'll also need a processor that discards messages for tombstoned records, since the main processor's trigger will not consume those messages due to it's check that filters out "missing records".

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    if (message.timestamp.getTime() === message.lastTimestamp?.getTime()){
+      // can discard if the last message is the same as the freshest message
+      if (message.lastValue === message.value){
+        return
+      }
+
+      const value = _.sortBy([message.value, message.lastValue])[0];
+
+      // can discard if the last message value > freshest message value
```
musnit, line 49:
```suggestion
      // can discard if the last message value > freshest message value, due to the string ordering rule
```

src/processors/default/processMempool.ts
```diff
@@ -0,0 +1,108 @@
...
+function categorizeMessages
+(
+  messages: PendingMempoolMessage[],
+):
+  { results: any, crdtUpdates: CrdtState[] } {
+  const entityUpdates: { [id: string]: any } = {};
+  const crdtUpdates: CrdtState[] = [];
+
+  // NOTE: messages can be partially processed. If a column in a message is stale, it will be ignored but the rest of the message
+  // will still be processed.
```
musnit, line 17:
```suggestion
  // NOTE: messages with stale columns are still fully processed. If a column in a message is stale, it will not be applied, but the message is still processed
```

Issue #69, opened 2022-09-01
make retry use Exponential backoff up to 2 hours


musnit:
we should have a longer backoff for retries. currently we'll try 5 times over 5 minutes. 5 times with backoff something like 1min, 2min, 5min, 10min, 30min, 90mins would be better. i guess retryCount^2.6 is probably fine to go with for the time?

this is a priority because sound.xyz sometimes does presales, where the api data is not available for 1 hour and so the mapper will error for an hour until the official release time. without exponential backoff, our retry mechanism will give up too soon to index these tracks

#79

PR #79, opened 2022-09-05
Feature - exponential backoff for error retries

puffo:
#69


src/triggers/errors.ts
```diff
@@ -7,7 +7,7 @@ export const errorRetry: Trigger<undefined> = async (clients) => {
   const errorQuery = `select *
   from "${Table.nftProcessErrors}"
   where ("numberOfRetries" < '${NUMBER_OF_RETRIES}' or "numberOfRetries" is null)
-  and (extract(minute from '${new Date().toISOString()}' - "lastRetry") >=1 or "lastRetry" is null)
```
puffo, line 10: I swapped `${new Date().toISOString()}` with `now()` so we use the database time. Any pitfalls I'm not seeing with using DB time over the nodejs process's time?

musnit: should be fine if it's the same type/format

src/triggers/errors.ts
```diff
@@ -7,7 +7,7 @@ export const errorRetry: Trigger<undefined> = async (clients) => {
...
+  and (age(now(), "lastRetry") >= make_interval(mins => cast(exp(coalesce("numberOfRetries", 0)) as int)) or "lastRetry" is null)
```
musnit, line 10: looks like this does e^numberOfRetries, with e meaning Euler's number?

so the retry intervals will be:
1 min, 2.7 mins, 7.2, 19.6, 53 ?

This doesn't get us past 60mins which i think will be necessary for some cases (like sound.xyz early-access drops, where they have a private 1hour mint period, during which we'd error since we cant access the data, and after which we'd process correctly)

I guess in terms of requirements, the important cases to cover are:
 - first retry after 1 minute
 - next retry after 5 mins (to cover the cases where ipfs propagation is a little slow
 - retry after 61 mins (to cover early private drops use case)
 
maybe we can just use pow("numberOfRetries", 2.8), ie to the power of 2.8, this gets us:
1, 2.8, 7.84, , 21.95, 61.4

puffo: Ah I was relying on us hitting 5 retries but I see we do `<` instead of `<=`.

Ok, so 2.8 doesn't get us where we want: { 1; 6,9; 21,6; 48,5; }

I'll adjust to use `pow("numberOfRetries", 3)`: { 1; 8; 27; 64; } 👍

musnit (approved):
cool looks good!

if you've done some manual QA yourself on this and you're happy, then good to merge :)

puffo:
Screenshot 2022-09-05 at 19 50 59

I've attached some examples and manually tested by modifying the timestamps and number of retries. All looks good to me!

PR #81, opened 2022-09-06
Generate migration for adding type-metadata to NFTs

puffo:
Preparing for changes in #76

### DIFF
--- a/src/db/migration-helpers.ts
+++ b/src/db/migration-helpers.ts
@@ -104,7 +104,7 @@ async function getForeignKeys(knex: Knex): Promise<{
   return result.rows
 }
 
-function tableNameToViewName(tableName: string): string {
+export function tableNameToViewName(tableName: string): string {
   // remove raw_ prefix
   return tableName.substring(4)
 }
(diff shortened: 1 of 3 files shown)


musnit:
ah sorry should have clarified this - i imagined we could just store the type metadata in the existing typeOverride json in each nftfactory record rather on an individual nft level, and just calculate it dynamically on the nft-level when we need it

eg:

overrides:
 ...
 nfts:
    1: {}
    2: {}
     ...

thoughts on that approach?

i think the motivation is that overriding on a per nft level is unlikely something we'll likely need to do manually/directly. it's primarily just to facilitate "passing the NFT along with the function that operates on it" as in https://github.com/spinamp/spindexer/pull/76#discussion_r962164421

puffo:
Hmmm, there are definitely trade offs with each approach. I think the benefits of this PRs suggested approach is nice from an encapsulation perspective but probably not worth all the redundant data. I will close this one out and try keep metadata on the NFTFactory 😄  Thanks for the input @musnit !

Some of the bigger impacts for each approach - definitely not an exhaustive list but maybe useful for a future Architecture Decision Record (ADR)

Copying across metadata directly to NFT:
-  ✅ function signatures for `mapNFTtoTrackID` and `mapNFTsToTrackIds` are cleaned up nicely and can operate without needing an extractor making them more self-contained. i.e. `mapNFTtoTrackID = (nft: NFT)` vs `mapNFTtoTrackID = (nft: NFT, extractor?: TitleExtractor)`
- ❗️lots of single use data stored on Nft table

Keeping metadata info on NFTFactory:
- ✅ no need for redundant data across multiple tables
- ❗️ more verbose function signatures and mixed responsibility for `mapNFTtoTrackID` and `mapNFTsToTrackIds`

PR #83, opened 2022-09-07
Clear track fixes

musnit:
@Geimaj after clearing out only track table, the ipfs media uploader skips uploads and just updates the table with the known cids that it already finds.

this works great, but is still slow due to the promise-per-minute limit. i think it's safe to increase this limit a ton so that it breezes through them all, since rate limiting is already handled sufficiently by the max requests limit.


Geimaj (approved):
LGTM!

PR #85, opened 2022-09-09
Add Mighty33 - Others Die tracks

puffo:


src/db/migrations/27-add-mighty-33-others-die.ts
```diff
@@ -0,0 +1,51 @@
...
+  contractType: NFTContractTypeName.default,
+  standard: NFTStandard.ERC721,
+  autoApprove: true,
+  approved: true,
+  typeMetadata: {
+    overrides: {
+      artist: {
+        name: 'Mighty33',
+        artistId: ethereumArtistId(mighty33Address),
+        avatarUrl: 'https://lh3.googleusercontent.com/IurC9VfibdFS8GtYHbPuEvMfDP_HD2YnLKFVLee9WouwIjtvQwFxBfc-UT6ypNRAAmK73Esfb92OyB3so4rXR3N-6_4iDhDElMzHPJQ=s0',
```
musnit, line 31:
```suggestion
        avatarUrl: 'https://web3-music-pipeline.mypinata.cloud/ipfs/QmS8dBxzgmNcCL71XYM5WxCkb2MM8rhQdCqAf1as8XXCVo',
```

src/db/migrations/28-add-mighty-33-others-die.ts
```diff
@@ -0,0 +1,54 @@
...
+import { addNftFactory, addPlatform, removeNftFactory, removePlatform } from '../migration-helpers';
+
+
+const mighty33Address = '0x8C62dD796e13aD389aD0bfDA44BB231D317Ef6C6';
+
+const MIGHTY_33: MusicPlatform = {
+  id: mighty33Address,
+  type: MusicPlatformType['multi-track-multiprint-contract'],
+  name: 'Mighty33',
+}
```
musnit, line 16: it seems like the platform is actualy https://geniuscorp.fr/ ? let's maybe keep it as that
```suggestion
const MIGHTY_33: MusicPlatform = {
  id: 'geniuscorp.fr',
  type: MusicPlatformType['multi-track-multiprint-contract'],
  name: 'geniuscorp.fr',
}
```

src/db/migrations/28-add-mighty-33-others-die.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const OTHERS_DIE: NftFactory = {
+  id: '0x97b9f21b41041e344f5bd71e3e86b69e79dcc0a6',
+  startingBlock: '10876005',
+  platformId: mighty33Address,
```
musnit, line 21:
```suggestion
  platformId: 'geniuscorp.fr',
```

src/db/migrations/28-add-mighty-33-others-die.ts
```diff
@@ -0,0 +1,54 @@
...
+        artistId: ethereumArtistId(mighty33Address),
+        avatarUrl: 'https://web3-music-pipeline.mypinata.cloud/ipfs/QmS8dBxzgmNcCL71XYM5WxCkb2MM8rhQdCqAf1as8XXCVo',
+        websiteUrl: 'https://geniuscorp.fr/m33'
+      },
+      track: {
+        websiteUrl: 'https://geniuscorp.fr/m33'
+      },
+      type: MusicPlatformType['multi-track-multiprint-contract'],
+      extractor: {
+        title: TitleExtractorTypes.METADATA_NAME,
```
musnit, line 39: i see the titles have track numbers in them like # 1, # 2 etc. i guess that's fine but may be nice to have an extractor that cuts those out, especially since it kinda looks bad with them in reverse album order

musnit: in future we'll have support for album/re-ordering but for now just that is probably good

src/types/fieldExtractor.ts
```diff
@@ -9,6 +10,7 @@ export type ExtractorTypes = {
 
 export enum TitleExtractorTypes {
   METADATA_NAME = 'metadata.name',
+  METADATA_NAME_WITHOUT_LEADING_INFO = 'metadataNameWithoutLeadingInfo',
```
puffo, line 13: I also considered `metadataNameWithoutTrackNumbers` but this seems slightly more explicit

src/utils/sanitizers.ts
```diff
@@ -0,0 +1,4 @@
+export const dropLeadingInfo = (str: string) => {
+  // remove everything before (and including) the first space
+  return str.replace(/^\S+\s+/g, '');
```
puffo, line 3: Also considered matching just the first few occurrences of the characters we don't want, but it ended up being a bit more complex to limit the matching to only the first ~5 characters. So I ended up going with a simpler approach to just drop everything before and with the first space.

The alternative (for reference)
```
  return str.replace(/[^a-zA-Z]{2,4}/g, '');
```

musnit (approved):

PR #87, opened 2022-09-09
Add primary constraint to views so that postgraphile generates byId endpoints for them

musnit:
@Geimaj We didn't have primary constraint comments on views, so postgraphile was not generating apis for things like processedTrackById.

This PR fixes that.


musnit:
merging this without review due to issues in production - @Geimaj plz review post-merge

PR #90, opened 2022-09-13
Feature - add hollyplus NFT collection

puffo:
- adds HollyPlus collection
- implements new `websiteUrl` extractor override to handle bad track URL metadata
- skipped dealing with the track artwork problems: `ERROR 9413: Could not resize the image: The animation is too large. ` not supported by pinata

holly-plus-qa

### DIFF
--- a/src/db/migrations/28-add-mighty-33-others-die.ts
+++ b/src/db/migrations/28-add-mighty-33-others-die.ts
@@ -1,6 +1,6 @@
 import { Knex } from 'knex';
 
-import { IdExtractorTypes, TitleExtractorTypes } from '../../types/fieldExtractor';
+import { IdExtractorTypes, TitleExtractorTypes, WebsiteUrlExtractorTypes } from '../../types/fieldExtractor';
 import { NftFactory, NFTContractTypeName, NFTStandard } from '../../types/nft';
 import { MusicPlatform, MusicPlatformType } from '../../types/platform';
 import { ethereumArtistId } from '../../utils/identifiers';
@@ -37,7 +37,8 @@ const OTHERS_DIE: NftFactory = {
       type: MusicPlatformType['multi-track-multiprint-contract'],
       extractor: {
         title: TitleExtractorTypes.METADATA_NAME_WITHOUT_LEADING_INFO,
-        id: IdExtractorTypes.USE_TITLE_EXTRACTOR
+        id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+        websiteUrl: WebsiteUrlExtractorTypes.METADATA_EXTERNAL_URL
       }
     }
   }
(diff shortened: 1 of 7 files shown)


src/types/fieldExtractor.ts
```diff
@@ -19,14 +20,38 @@ export enum IdExtractorTypes {
...
+const useTokenIdAppendedExternalUrl = (nft: NFT): string => {
+  const url = new URL(nft.metadata.external_url);
+  return `${url.origin}/token/${nft.tokenId}`;
+}
+
+export const websiteUrlExtractor = (contract: NftFactory): Extractor => {
+  const websiteUrlExtractorOverride = contract.typeMetadata?.overrides?.extractor?.websiteUrl;
+  if (!websiteUrlExtractorOverride) {
+    throw new Error('unknown extractor override provided')
+  }
```
musnit, line 53: I think for the websiteUrlExtractor it's fine to default to WebsiteUrlExtractorTypes.METADATA_EXTERNAL_URL if none exists, since that's very much a common standardized thing that is fair as a sane default.

this also means we can avoid doing all the past migration fixes

src/db/migrations/28-add-mighty-33-others-die.ts
```diff
@@ -37,7 +37,8 @@ const OTHERS_DIE: NftFactory = {
       type: MusicPlatformType['multi-track-multiprint-contract'],
       extractor: {
         title: TitleExtractorTypes.METADATA_NAME_WITHOUT_LEADING_INFO,
-        id: IdExtractorTypes.USE_TITLE_EXTRACTOR
+        id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+        websiteUrl: WebsiteUrlExtractorTypes.METADATA_EXTERNAL_URL
```
musnit, line 41: As you mentioned a while back, we should stop changing past migrations as that doesn't really work unless we rebuild the whole DB. We'd need to do these changes in migration 29 ideally.

(But, I don't think these migrations are necessary anyway given my other comment on default urls)

puffo: Cool, makes sense

src/db/migrations/29-add-holly-plus.ts
```diff
@@ -0,0 +1,46 @@
...
+    overrides: {
+      artist: {
+        name: 'hollyplus',
+        websiteUrl: 'https://auction.holly.plus/',
+      },
+      type: MusicPlatformType['multi-track-multiprint-contract'],
+      extractor: {
+        title: TitleExtractorTypes.METADATA_NAME,
+        id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+        websiteUrl: WebsiteUrlExtractorTypes.USE_TOKEN_ID_APPENDED_EXTERNAL_URL,
```
musnit, line 32: This is nice! Good example of doing the extra step/extra QA to get the best data

puffo:
Approval via DMs 👍

PR #94, opened 2022-09-14
Add tables and migration for collectors

puffo:
Extracts "owner" data to a new many-to-many relationship via `NftsCollectors` and `Collectors`.

- retains the old "owner" column for backward compatibility until we update the frontend to use the newer data (staged release)

Also adds "burned" data to the `Nfts` table, opening up possibilities to skip showing any tracks that have been burned.


src/db/db.ts
```diff
@@ -5,19 +5,21 @@ import { Cursor } from '../types/trigger'
...
-  processors = 'raw_processors',
+  collectors = 'raw_collectors',
+  erc721Transfers = 'raw_erc721_transfers',
+  ipfsPins = 'raw_ipfs_pins',
+  ipfsFiles = 'raw_ipfs_files',
   metaFactories = 'raw_meta_factories',
-  nftFactories = 'raw_nft_factories',
+  nfts = 'raw_nfts',
   nfts_processedTracks = 'raw_nfts_processed_tracks',
+  nftsCollectors = 'raw_nfts_collectors',
```
puffo, line 17: Just reordered this and added `nftsCollectors` and `collectors`

src/db/migration-helpers.ts
```diff
@@ -139,6 +139,15 @@ export async function updateViews(knex: Knex){
 
   // create views
   for (const table of tables) {
+
+    // return early if table doesn't exist
```
puffo, line 143: To get us closer towards using migrations properly, this was added to make it more robust to new values being added in `db/tables`

src/db/migrations/30-add-collector-table.ts
```diff
@@ -0,0 +1,57 @@
...
+import { tableNameToViewName, updateViews } from '../migration-helpers';
+
+export const up = async (knex: Knex) => {
+  await knex.schema.createTable(Table.collectors, table => {
+    table.string('id').notNullable()
+    table.primary(['id'])
+  } )
+
+  await knex.schema.createTable(Table.nftsCollectors, table => {
+    table.string('nftId') //.references('id').inTable(Table.nfts).onDelete('cascade'); // cannot enforce this because we might not have the nft yet due to order of processing
```
puffo, line 14: Suggestions for how to get around this order dependency? It is currently problematic because we deal with transfers ahead of populating any NFTs

src/db/migrations/30-add-collector-table.ts
```diff
@@ -0,0 +1,57 @@
...
+    FROM ${Table.nfts}
+    WHERE owner IS NOT null;
+  `)
+
+  await knex.raw(`
+    INSERT INTO ${Table.nftsCollectors} ("nftId", "collectorId")
+    SELECT DISTINCT id, owner
+    FROM ${Table.nfts}
+    WHERE owner IS NOT null;
+  `)
```
puffo, line 33: Will probably remove this and rather reindex when we're closer to merging

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
 
+      const contractAddress = formatAddress(contract.id);
+      const fromAddress = formatAddress(item.args!.from);
+      const toAddress = formatAddress(item.args!.to);
       const tokenId = BigInt((item.args!.tokenId as BigNumber).toString());
-      const newMint = item.args!.from === ETHEREUM_NULL_ADDRESS;
+      const nftId = contractType.buildNFTId(contractAddress, tokenId);
+
       transfers.push({
-        id: `${CHAIN}/${item.blockNumber}/${item.logIndex}`,
```
puffo, line 37: Just some straightforward cleanup on this file

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
-          id: contractType.buildNFTId(contract.id, tokenId),
-          owner: item.args!.to
+
+      if (!newMint(fromAddress)) {
+        updatedNFTs.push({
+          id: nftId,
+          owner: toAddress
         })
+        updatedNftsCollectors.push({ nftId: nftId, collectorId: fromAddress, amount: -1 })
+        updatedNftsCollectors.push({ nftId: nftId, collectorId: toAddress, amount: 1 })
```
puffo, line 58: See other comment in this file for the overall strategy.

But here, we register a transfer as  +1 amount for receiver, -1 amount for sender

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
         tokenId,
         platformId: contract.platformId,
-        owner: item.args!.to,
+        owner: toAddress,
         approved: contract.autoApprove
       });
+      updatedNftsCollectors.push({ nftId: nftId, collectorId: toAddress, amount: 1 })
     });
 
+    // update Collectors table
```
puffo, line 73: The overall strategy is to effectively:
1. collect all transfers and assign to collectors  (+1 amount for receiver, -1 amount for sender)
2. add any mints (+1 for receiver)
3. update DB with any missing collector addresses
4. update/add any NftsCollectors with the appropriate amounts

Very brute force but I think it can work with a bit of cleanup and optimization

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
         createdAtEthereumBlockNumber: '' + item.blockNumber,
-        contractAddress: formatAddress(contract.id),
+        contractAddress: contractAddress,
         tokenId,
         platformId: contract.platformId,
-        owner: item.args!.to,
+        owner: toAddress,
         approved: contract.autoApprove
       });
+      updatedNftsCollectors.push({ nftId: nftId, collectorId: toAddress, amount: 1 })
```
puffo, line 70: Minted NFT -> +1 to receiver

src/types/nftsCollectors.ts
```diff
@@ -0,0 +1,23 @@
...
+  const grouped = _.groupBy(allNftsCollectors, (e) => { return `${e.collectorId}-${e.nftId}` })
+  const consolidatedGroups = _.mapValues(grouped, (v) => {
+    return {
+      nftId: v[0].nftId,
+      collectorId: v[0].collectorId,
+      amount: _.sumBy(v, (el) => { return el.amount } )
+    }
+  })
+  return Object.values(consolidatedGroups)
+}
```
puffo, line 23: Could definitely use some input on how to do this more cleanly/efficiently

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
musnit: i think it looks alright - i see the stronger motivation for separating this out into a different processor now as u suggested, since that would simplify it further but the current implementation still looks good so i think we can keep it.

suggestions:
 - move line 75 down, ie, the Table.collectors insert so that we do all the inserts together instead of interweaving logic/getRecords and inserts
 - could you move both the Table.collectors insert and Table.nftsCollectors inserts down below the newNFTs insert to solve your ordering problem related to the foreignkey on the nftCollectors table?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
         throw 'buildNFTId not specified'
       }
 
+      const contractAddress = formatAddress(contract.id);
```
musnit, line 35: hmm i think we weren't formatting the address before right?

i guess this means that if someone is following a collector today, they are following an unformatted address and when they change to formatted addresses, they may lose their follows?

i'm not sure the formatting/api from the frontend, so haven't doubled check if this will actually happen, but I think it's not a big deal so I don't think we should worry about handling it.

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
       });
-      if (!newMint) {
-        updates.push({
-          id: contractType.buildNFTId(contract.id, tokenId),
-          owner: item.args!.to
+
+      if (!newMint(fromAddress)) {
+        updatedNFTs.push({
+          id: nftId,
+          owner: toAddress
```
musnit, line 55: do we still need this? this is just for the staged migration/backwards compatibility right?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,43 +32,68 @@ const processorFunction = (contracts: NftFactory[]) =>
...
+    // fetch any existing NftCollectors
+    const idPairs = _.uniq(updatedNftsCollectors.map( (e) => [e.nftId, e.collectorId] ))
+    const existingNftsCollectors = await clients.db.getRecords<NFTsCollectors>(Table.nftsCollectors,
+      [ ['whereIn', [ ['nftId', 'collectorId'], idPairs ] ] ]
+    );
+
+    // update all NFTsCollectors with most up to date amounts
+    const fullNftsCollectors = updatedNftsCollectors.concat(existingNftsCollectors)
+    const consolidatedNftsCollectors = consolidate(fullNftsCollectors)
+    await clients.db.insert(Table.nftsCollectors, consolidatedNftsCollectors, { ignoreConflict: ['nftId', 'collectorId'] });
```
musnit, line 86: this all looks good. just a small suggestion to change variable naming for ease of readibility/understanding:
 - updatedNftsCollectors is really a set of changes, rather than a set of records, so even though in this case the data type for those changes has the same fields and structure as the data type for the underlying record, they are a little conceptually different
 - i'd create a seperate type to make this explicit, like, `type NFTsCollectorChanges = NFTsCollectors` 
 - instead `updatedNftsCollectors: NFTsCollectors[]` i'd name it something like `nftCollectorChanges: NFTsCollectorChanges[]` and fullNftsCollectorChanges
 - and then just use the name updatedNftsCollectors instead of consolidatedNftsCollectors

src/db/migrations/30-add-collector-table.ts
musnit: i think the processor should always be inserting nfts before transfers & nftsCollectors, as in the later comment to solve this?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
musnit: you're already doing this above for the !newMint scenario, so won't this happen twice in that scenario? should this line be in an else in the above conditional?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: Ok, so just to clarify, should we keep as is or revert to leaving it unformatted?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: Correct, just needed for compatibility.

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: Ah, I thought the `return undefined` on 59 was breaking us out of the loop?

I've moved it to an `else` statement for readability either way.

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: I've colocated the inserts. Also tried moved those two lower down in the order but oddly enough it doesn't fix the problem 😕 

I think there may be some conditions where the above NFTs are filtered out before the insert/upsert. All I can see is  line 89
```ts
await clients.db.insert(Table.nfts, newNFTs.filter(n => !!n), { ignoreConflict: 'id' });
```
But that's just clearing empty ones, no? Am I missing something elsewhere?

musnit: yeah that is weird. maybe something is wrong/may need to debug which things are failing and why

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
musnit: oh yeah i missed that, sorry

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
musnit: keep as is

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: Yeaaahhh! Much nicer!!!! 💯

src/db/migrations/28-add-mighty-33-others-die.ts
```diff
@@ -17,7 +17,7 @@ const MIGHTY_33: MusicPlatform = {
 
 const OTHERS_DIE: NftFactory = {
   id: '0x97b9f21b41041e344f5bd71e3e86b69e79dcc0a6',
-  startingBlock: '10876005',
+  startingBlock: '10876004',
```
musnit, line 20: hm looks like this contract was actually created on block 10688108?

See https://etherscan.io/tx/0x69a998a7b1290d015b84682d727b70f686b8277297b902a1f7112c717820d48a

Also, I guess since we won't pickup this past change automatically, in order to actually deploy this we'll need to either:
 - delete/drop this NFTFactory and all it's cascading nfts/tracks as part of a new migration
 - delete/drop this NFTFactory and all it's cascading nfts/tracks just manually when we deploy
 - rebuild whole db from scratch and replace with new rebuild

lmk whichever you think is quickest/easiest

src/types/ethereum.ts
```diff
@@ -1 +1,40 @@
...
+  '0x8888888888888888888888888888888888888888',
+  '0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb',
+  '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee',
+  '0xffffffffffffffffffffffffffffffffffffffff',
+  '0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+  '0x1234567890123456789012345678901234567890',
+  '0x0123456789012345678901234567890123456789',
+  '0xdead000000000000000042069420694206942069',
+  '0x000000000000000000000000000000000000dead',
+]
```
musnit, line 32:
```suggestion
  '0x000000000000000000000000000000000000dead',
  '0x000000000000000000000000000000000000d34d',
]
```

src/db/migrations/28-add-mighty-33-others-die.ts
puffo: Oh damn! I've been getting the starting block based on the earliest transaction that's taken place within that contract, i.e. https://etherscan.io/txs?a=0x97b9f21b41041e344f5bd71e3e86b69e79dcc0a6

Moving forward I'll make sure to look at the contract inception block (and add it to my guide).

puffo: RE: deployment
I think we should do a full rebuild given that we want to pick up the proper +1 and -1's for each transfer between collectors, for all contracts on the system since the beginning of indexing.

I did a full rebuild on the QA environment earlier, so I think we can just dump that and update the repo?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,38 +32,60 @@ const processorFunction = (contracts: NftFactory[]) =>
...
+    );
+
+    const allNftsCollectorsChanges = nftsCollectorsChanges.concat(existingNftsCollectors)
+    const updatedNftsCollectors = consolidate(allNftsCollectorsChanges)
+
     await clients.db.insert(Table.nfts, newNFTs.filter(n => !!n), { ignoreConflict: 'id' });
-    await clients.db.update(Table.nfts, updates);
+    await clients.db.upsert(Table.nfts, updatedNFTs);
+    await clients.db.insert(Table.collectors, _.uniqBy(allCollectors, 'id'), { ignoreConflict: 'id' })
+    await clients.db.insert(Table.nftsCollectors, updatedNftsCollectors, { ignoreConflict: ['nftId', 'collectorId'] });
```
musnit, line 88: these are both basically upserts right? should we just use our upsert function if it is capable of doing the same thing?

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -31,38 +32,60 @@ const processorFunction = (contracts: NftFactory[]) =>
...
+    const existingNftsCollectors = await clients.db.getRecords<NFTsCollectors>(Table.nftsCollectors,
+      [ ['whereIn', [ ['nftId', 'collectorId'], idPairs ] ] ]
+    );
+
+    const allNftsCollectorsChanges = nftsCollectorsChanges.concat(existingNftsCollectors)
+    const updatedNftsCollectors = consolidate(allNftsCollectorsChanges)
+
     await clients.db.insert(Table.nfts, newNFTs.filter(n => !!n), { ignoreConflict: 'id' });
-    await clients.db.update(Table.nfts, updates);
+    await clients.db.upsert(Table.nfts, updatedNFTs);
```
musnit, line 86: Still don't follow the scenario in which it's possible for updatedNFTs to include an nft that does not yet exist, so don't understand the need for an upsert here. Can you clarify?

src/db/migrations/28-add-mighty-33-others-die.ts
musnit: sounds good

musnit: > Oh damn! I've been getting the starting block based on the earliest transaction that's taken place within that contract, i.e. https://etherscan.io/txs?a=0x97b9f21b41041e344f5bd71e3e86b69e79dcc0a6
> 
> Moving forward I'll make sure to look at the contract inception block (and add it to my guide).

yep exactly - we should use the deployment block

puffo: Ok, I've double checked the previous contracts I had added and they're all correct. Will push up the fix for this block along with the other changes.

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: Yep, should be the same result. I'll switch to `upsert`

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
puffo: The culprit was an [off by one error](https://www.martinfowler.com/bliki/TwoHardThings.html) with a starting block for the ZoraOriginal contract. Resolved!

src/db/migrations/05-add-songcamp-chaos.ts
```diff
@@ -4,15 +4,15 @@ import { NftFactory, NFTContractTypeName, NFTStandard } from '../../types/nft';
...
+const CHAOS_PLATFORM: MusicPlatform = {
   id: '0x8427e46826a520b1264b55f31fcb5ddfdc31e349',
   type: MusicPlatformType.chaos,
   name: 'Chaos',
 }
 
 const CHAOS_CONTRACT: NftFactory = {
   id: '0x8427e46826a520b1264b55f31fcb5ddfdc31e349',
-  startingBlock: '10766312',
+  startingBlock: '10766312', // TODO: check this
```
musnit, line 15: @puffo still todo?

src/db/migrations/05-add-songcamp-chaos.ts
```diff
@@ -4,15 +4,15 @@ import { NftFactory, NFTContractTypeName, NFTStandard } from '../../types/nft';
...
+const CHAOS_PLATFORM: MusicPlatform = {
   id: '0x8427e46826a520b1264b55f31fcb5ddfdc31e349',
   type: MusicPlatformType.chaos,
   name: 'Chaos',
 }
 
 const CHAOS_CONTRACT: NftFactory = {
   id: '0x8427e46826a520b1264b55f31fcb5ddfdc31e349',
-  startingBlock: '10766312',
+  startingBlock: '14891074'
```
musnit, line 15:
```suggestion
  startingBlock: '14891073'
```

should be -1

Issue #95, opened 2022-09-17
replace blocks subgraph with custom block processor


Geimaj:
remove dependency from the graph and query ethereum blocks directly from our eth node

PR #96, opened 2022-09-17
replace subgraph query with rpc queries

Geimaj:
Implements #95 to prepare for the sunsetting of the graph hosted service

### DIFF
--- a/src/processors/default/addTimestampToERC721NFTs.ts
+++ b/src/processors/default/addTimestampToERC721NFTs.ts
@@ -1,3 +1,4 @@
+
 import _ from 'lodash';
 
 import { Table } from '../../db/db';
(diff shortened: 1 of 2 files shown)


src/processors/default/addTimestampToERC721NFTs.ts
```diff
@@ -7,13 +8,14 @@ import { Record } from '../../types/record';
 
 const processorFunction = (table: Table) => async (items: Record, clients: Clients) => {
   const recordsByBlockNumber = _.groupBy(items, 'createdAtEthereumBlockNumber');
-  const blockNumbers = Object.keys(recordsByBlockNumber);
+  const blockNumbers = Object.keys(recordsByBlockNumber).map(blockNumber => Number.parseInt(blockNumber));
   console.log(`Processing until block ${blockNumbers[blockNumbers.length - 1]}`)
-  const timestampsResponse = await clients.blocks.fetchBlockTimestamps(blockNumbers);
```
puffo, line 12: Looks like we're not using `fetchBlockTimestamps` anywhere else, so we can likely remove the BlocksClient from the list of `Clients`, and then also nuke `blocks.ts` and the imports associated with it.

Also the `BLOCKS_SUBGRAPH_ENDPOINT` goes away! One less external dependency!

musnit:
Would be nice to remove the graph dependency entirely and use this, but I am concerned that doing so will slow down the indexer considerably since we now need an additional HTTP request for every single ethereum block that contains an nft?

With the graph, we at least could do like 1000 requests in bulk.

Have you tried this PR and tested the impact it has on performance?

Some alternatives we could consider if performance is heavily impacted:
 a - Setup and run our own isolated graph node as part of the Spindexer stack, so there is no longer any third party dependency
 b - Look into using batch eth RPC requests (https://geth.ethereum.org/docs/rpc/batch) assuming our providers support it
 c - Modify our database so that we cache block timestamps in our own DB, and fix the backup/reset process and scripts so that even when they do a full db reset, the timestamps backup is used so that we don't need to re-index old timestamps

If (b) works, then probably the easiest to do now and should fit well into this existing PR, so maybe try that first?

Geimaj:
Yeah it does slow down things quite a bit since we have to do a request for each block. Fetching the timestamps for each block from initial sync takes at least 20 min from my testing so far. 

According to the alchemy docs: 
`Generally speaking, Alchemy does not recommend using batch requests as they can be less reliable compared to individual API calls.

Alchemy does not recommend using batch requests for any methods. These requests tend to be less performant than sending them individually, so if you’re trying to optimize speed, this might not be the best option.`

So I don't think (b) would be the right option here. Let's go with (a) for now and defer (c) then

musnit:
@Geimaj our node is up. will take a day or so to catch up to latest ethereum block - will migrate over when its done

PR #99, opened 2022-09-19
Add RELICS artist migration

puffo:
- new AudioUrl, ArtistId, and ArtistName metadata overrides


src/db/migrations/31-add-relics-season-1.ts
```diff
@@ -0,0 +1,48 @@
...
+      artist: {
+        // TODO: add extractor for artist ID?
+        name: 'RELICS',
+        websiteUrl: 'https://relics.xyz',
+        avatarUrl: 'https://web3-music-pipeline.mypinata.cloud/ipfs/QmS6WbpHrMYsu6Tmts61FfqiEwmrzJRrPrnN9o5qGDzME4',
+      },
+      type: MusicPlatformType['multi-track-multiprint-contract'],
+      extractor: {
+        id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+        title: TitleExtractorTypes.METADATA_NAME,
```
musnit, line 33: side comment - should we consider making both of these a default, rather than required?

src/db/migrations/31-add-relics-season-1.ts
```diff
@@ -0,0 +1,48 @@
...
+  startingBlock: '14180427',
+  platformId: 'relics',
+  contractType: NFTContractTypeName.default,
+  standard: NFTStandard.ERC721,
+  autoApprove: true,
+  approved: true,
+  typeMetadata: {
+    overrides: {
+      artist: {
+        // TODO: add extractor for artist ID?
```
musnit, line 25: yeah, there doesn't seem to be any great way to do it with relics yet, like an artist address or such, but I do think we should separate artists for relic in this PR nonetheless.

i guess best option for relics atm looks like:
 - artistNameExtractor that gets the Musician trait
 - artistIdExtractor that returns something like `${nftFactory.platformId}/${artistNameExtractor(nft)}`
 - websiteURL just https://relics.xyz/ for all for now?
 - avatarUrl all the same for now, as is?

src/db/migrations/31-add-relics-season-1.ts
```diff
@@ -0,0 +1,48 @@
...
+        name: 'RELICS',
+        websiteUrl: 'https://relics.xyz',
+        avatarUrl: 'https://web3-music-pipeline.mypinata.cloud/ipfs/QmS6WbpHrMYsu6Tmts61FfqiEwmrzJRrPrnN9o5qGDzME4',
```
musnit, line 28: this link doesn't open an image properly - do you know what filetype this actually is?

puffo: It's a .avif file, I'll do a quick conversion and replace it with .jpg

puffo: I've just tried it with local spinamp instance and the image displays. I think we should leave it seeing as it works in the browser, or not?

puffo:
Track thumbnails are processed in the backend but pinata gives the following error when trying to load them on the frontend. It looks like the artwork is the full track with video media. 

`https://spinamp.mypinata.cloud/ipfs/QmbXrJKLLKoWpw4Q746snpgKnvfnUeuaHRKN67iSHMD3Kz?img-width=350&img-fit=cover`

Status: 415 Response: 
Unsupported Media Type - ERR_ID:00021

Any suggestions @musnit @Geimaj ?

musnit:
> Track thumbnails are processed in the backend but pinata gives the following error when trying to load them on the frontend. It looks like the artwork is the full track with video media.
> 
> `https://spinamp.mypinata.cloud/ipfs/QmbXrJKLLKoWpw4Q746snpgKnvfnUeuaHRKN67iSHMD3Kz?img-width=350&img-fit=cover`
> 
> Status: 415 Response: Unsupported Media Type - ERR_ID:00021
> 
> Any suggestions @musnit @Geimaj ?

ah yeah :S. pinata kinda still sucks - we could potentially migrate the frontend to use cloudinary endpoints for images too now, instead of pinata - similar to how it does with audio. though seems to have problems too: https://snowfork.mo.cloudinary.net/QmbXrJKLLKoWpw4Q746snpgKnvfnUeuaHRKN67iSHMD3Kz

src/db/migrations/31-add-relics-season-1.ts
puffo: I think it is ok to leave for now. Extracting all these migrations into CRDT seeds will likely guide us for what kind of extractors cleanup is required

puffo:
Should we forge ahead without track thumbnails to just get the tracks into spindexer?

src/types/fieldExtractor.ts
```diff
@@ -81,3 +126,19 @@ export const resolveEthereumTrackIdOverrides = (nft: NFT, contract: NftFactory):
...
+    return contract.platformId;
+  }
+  return artistNameExtractor(contract)(nft);
+}
+
+export const resolveArtistIdOverrides = (nft: NFT, contract: NftFactory): string => {
+  const artistIdExtractorOverride = contract.typeMetadata?.overrides?.extractor?.artistId;
+  if (!artistIdExtractorOverride || (artistIdExtractorOverride !== ArtistIdExtractorTypes.USE_PLATFORM_AND_ARTIST_NAME)) {
+    return contract.platformId;
+  }
```
musnit, line 142: hm although this was the default before, i don't think it's a good default. i think we should just hard error here and make sure the legacy migrations include a USE_PLATFORM_ID or similar extractor

src/types/platforms-types/multi-track-multiprint-contract.ts
```diff
@@ -59,9 +59,10 @@ const mapArtistProfile = ({ apiTrack, nft, contract }: { apiTrack: any, nft?: NF
   if (!contract) {
     throw new Error(`Contract missing for mapArtistProfile for nft ${nft.id}`)
   }
+
   return {
-    name: contract.platformId,
-    artistId: contract.platformId,
+    name: resolveArtistNameOverrides(nft, contract),
+    artistId: resolveArtistIdOverrides(nft, contract),
```
musnit, line 65: comment on function naming: here we use resolveArtistNameOverrides whereas the others just use titleExtractor or xyzExtractor. should we use artistNameExtractor and artistIdExtractor for the function names here to be more consistent, or is there something different going on here that lead you to use a different naming convention?

src/types/fieldExtractor.ts
musnit: we can also just manually update the legacy changed NftFactories so we don't have to do a rebuild

src/types/platforms-types/multi-track-multiprint-contract.ts
puffo: It is unfortunately a bit more nuanced. 

The extractors only take in an `NftFactory` to extract which strategy should be used when making/deriving another field. i.e. 
```ts
audioUrlExtractor = (contract: NftFactory): Extractor => { ... }
// where Extrator is:
type Extractor = (nft: NFT) => string;
```

The "resolvers" take both the `NftFactory` and the `NFT` to then automatically execute the strategy on a particular nft. i.e.

```ts
export const resolveArtistNameOverrides = (nft: NFT, contract: NftFactory): string => { ... }
```

Agree that the naming isn't doing a good job of indicating that. 
I think this file needs some care after it settles down and there's less churn.

src/types/fieldExtractor.ts
puffo: I've changed this and am reviewing old migrations.

Question @musnit : if we change previous ones to `USE_PLATFORM_AND_ARTIST_NAME`, will that break any fav'd artists for existing users?

musnit: probably yes but may not be a big deal.

having said that, ideally i don't think we should ever really be using platforms as part of the artist identity, since they will often exist across multiple platforms, so i wouldn't change to that. it's just a band-aid for relics for now i think

puffo: Ok, having something like `ARTIST_NAME` probably makes more sense.

puffo: I've made it fall back to the artist ID override, and it throws an error if no extract is provided and there's no artist ID override.

src/db/migrations/29-add-holly-plus.ts
```diff
@@ -23,6 +23,7 @@ const HOLLY_PLUS: NftFactory = {
     overrides: {
       artist: {
         name: 'hollyplus',
+        artistId: 'hollyplus',
```
musnit, line 26: why is this needed?

puffo: It was previously falling back to platform id, which was `hollyplus`.

Now that we've removed that bad default, I've added this to be explicit about what was here before.

src/types/fieldExtractor.ts
```diff
@@ -81,3 +134,43 @@ export const resolveEthereumTrackIdOverrides = (nft: NFT, contract: NftFactory):
...
+    const url = (track as any).lossyArtworkURL;
+    if (!url) {
+      throw new Error('track lossy artwork url override not provided');
+    }
+    return url;
+  }
+  return nft.metadata.image
+}
+
+export const resolveArtistNameOverrides = (nft: NFT, contract: NftFactory): string => {
```
musnit, line 152: i don't see where overrides.artist.name is being used ever? is it no longer used? eg: on danial allen, we specify overrides.artist.name but never specify overrides.extractor and so it doesn't seem like the name will ever be used. did you try recreate daniel allen/check this? i may be missing something, but at the moment it looks like the name will be wrong if we recreate?

musnit: the resolveArtistIdOverrides logic looks good though - should this one look more like that?

src/types/fieldExtractor.ts
```diff
@@ -81,3 +134,43 @@ export const resolveEthereumTrackIdOverrides = (nft: NFT, contract: NftFactory):
...
+      return contract.platformId;
+    }
+    else if (artistIdExtractorOverride === ArtistIdExtractorTypes.USE_PLATFORM_AND_ARTIST_NAME) {
+      return `${contract.platformId}/${slugify(resolveArtistNameOverrides(nft, contract))}`;
+    }
+  }
+
+  const artistIdOverride = contract.typeMetadata?.overrides?.artist?.artistId;
+  if (!artistIdOverride) {
+    throw new Error('must directly provided an artist ID override, or an artist ID extractor override');
```
musnit, line 173: should we default to platformId as before? or otherwise i think this may break chaos so we may need to add an artist id override on chaos too

src/db/migrations/29-add-holly-plus.ts
musnit: also should we use ethereumArtistId(address) if we do need this? or maybe make ethereumArtistId(nftFactoryAddress) a default that everything uses

musnit: > It was previously falling back to platform id, which was `hollyplus`.
> 
> Now that we've removed that bad default, I've added this to be explicit about what was here before.

ah ok - well since hollyplus hasn't actually been released/deployed yet, maybe the better default of ethereumArtistId(nftFactoryAddress) is fine for it?

src/types/fieldExtractor.ts
musnit: re chaos - nvm, i see it doesn't actually use multi-track-multiprint, i thought it did

src/types/fieldExtractor.ts
puffo: They are still being used as the highest priority override by being destructured, eg in the `mapArtistProfile` method:
```ts
return {
    // ...
    createdAtEthereumBlockNumber: nft.createdAtEthereumBlockNumber,
    ...contract.typeMetadata?.overrides?.artist
}
```

puffo: Given that we're shifting the fallback for artist id to be `ethereumArtistId(nftFactoryAddress)`, I'll make an explicit `ArtistIdExtractorTypes` ->`USE_ARTIST_ID_OVERRIDE = 'useArtistIdOverride',` and use it explicitly where needed.

puffo: > the resolveArtistIdOverrides logic flow looks good though - should this one look more like that?

I've applied a similar strategy for ArtistName and removed the fallback to platformId as I think it isn't a good default. You now have to explicitly specify how to deal with the artist name

PR #104, opened 2022-09-27
better error logging and support null ids

musnit:


src/types/fieldExtractor.ts
```diff
@@ -6,6 +6,7 @@ import { getTrait, NFT, NftFactory } from './nft';
 type Extractor = (nft: NFT) => string;
 type StrategyExtractor = (nftFactory: NftFactory) => Extractor;
 type Resolver = (nft: NFT, contract: NftFactory) => string;
+type IDResolver = (nft: NFT, contract: NftFactory) => string | null;
```
puffo, line 9: Slightly odd to have a null ID but I think it's something to tackle at a later stage.

puffo (approved):

PR #105, opened 2022-09-27
Pass sound xyz credentials for API calls

puffo:


puffo:
Tested locally, approved via discord.

PR #107, opened 2022-09-27
Extract migrations as CRDT seeds and persist `db/seeds.dump`

puffo:
- refactors all migrations containing platform/factory/metafactory information into `artistIntegrations.ts`
- loads data into DB via CRDT seeds in migration 38
- changes the `value` on `crdt_state` from varchar(255) to varchar(1000) as some bigger typeMetadata instructions needed it
- adds yarn commands for saving, clearing, and restoring seeds table


musnit:
Liking good. Do u think it would also make sense to:
 - save seeds table in backup
 - remove migrations for seeds entirely and just insert the seeds via our Postgres backup tables with something like yarn restore-seeds
 - do rebuild with clean migration state

?

I guess only thing may b that we may wanna use plaintext pg_dump backups for the seeds table instead of compressed ones so that they’re easier to read and have nicer diffs on GitHub when we save new versions

puffo:
Yeah, plaintext seeds table dump would make a lot of sense. 

The existing compressed DB dumps might get out of sync with the seeds table dump but it probably isn't a huge problem given that we can always do a full rebuild and come to the same end state eventually.

puffo:
> remove migrations for seeds entirely

Yes to removing and renumbering the migrations, but I do think we can use a handful (if not all) of the `artistMigrations.ts` to define some kind of acceptable API docs for the seeds API's payloads. 

So I'd like to keep those in for a little while longer. Otherwise we should at least keep track of a commit for us to reference later on

musnit:
> The existing compressed DB dumps might get out of sync with the seeds table dump but it probably isn't a huge problem given that we can always do a full rebuild and come to the same end state eventually.

yeh sounds ok

src/db/migrations/37-change-crdt-state-table-value-column-capacity.ts
```diff
@@ -0,0 +1,20 @@
+import { Knex } from 'knex';
+
+import { Table } from '../db';
+import { tableNameToViewName, updateViews } from '../migration-helpers';
+
+export const up = async (knex: Knex) => {
+  await knex.schema.dropView(tableNameToViewName(Table.crdtState));
+  await knex.schema.alterTable(Table.crdtState, table => {
+    table.string('value', 1000).alter();
```
musnit, line 9: oh nice catch. should this maybe be even longer/bigger? it looks like the biggest we have in practice that could potentially be exposed is 20000, so i think going with that should be fine

musnit (approved):
one small comment, but otherwise looks good, thanks!

do you think we should do a full rebuild? if so, then maybe we can just delete all the migrations instead of deprecating them?

PR #109, opened 2022-09-29
Seeds API V1

puffo:


src/utils/seedsApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    next();
+  } catch (e) {
+    response.status(403).send('Authentication failed');
+    return;
+  }
+};
+
+export function validateSignature(signatureData: {
```
Geimaj, line 36:
```suggestion
export function validateSignature(signatureData: {
message: string;
signature: string;
},
permissionedAddresses: string[]
```

src/utils/seedsApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    !signatureData.signature
+  ) {
+    console.error('No signature provided');
+    throw 'No signature provided';
+  }
+
+  if (
+    !isValidChecksumAddress(signer) ||
+    // TODO: change to only allow a single ENV-stored signer
+    signatureData.signer.toLocaleLowerCase() !== signer.toLocaleLowerCase()
```
Geimaj, line 63:
```suggestion
    !permissionedAddresses.includes(signatureData.signer.toLowerCase())
```

test/serve_seeds_api.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+  })
+
+  describe('authenticated', () => {
+    describe('platforms', () => {
+      const platformsEndpoint = '/v1/seeds/platforms'
+
+      describe('with the incorrect shape', () => {
+        it('returns an error', (done) => {
+          const message = wallet.sign(JSON.stringify({ blam: 'yam' }))
```
Geimaj, line 36:
```suggestion
          const body = { blam: 'yam' }
          const signature = wallet.sign(JSON.stringify(body)).signature
```

test/serve_seeds_api.ts
```diff
@@ -0,0 +1,65 @@
...
+          const body = { msg: message, sig: message.signature, address: wallet.address }
+
+          supertest(app).post(platformsEndpoint).send(body)
```
Geimaj, line 39:
```suggestion
          supertest(app).post(platformsEndpoint).send(body, {
            headers: {
              sig: signature
            }
          })
```

src/serve_seeds_api.ts
```diff
@@ -0,0 +1,36 @@
...
+
+app.post(`${apiVersionPrefix}/seeds/`, async (req, res) => {
+  try {
+    validateSeed(req.body)
+  } catch (e: any) {
+    return res.status(422).send({ error: e.message });
+  }
+
+  try {
+    persistSeed(req.body)
```
musnit, line 23:
```suggestion
    await persistSeed(req.body)
```

src/utils/seedsApi.ts
```diff
@@ -0,0 +1,145 @@
...
+    } else if (payload.entity === 'contract') {
+      message = getCrdtUpdateMessage<NftFactory>(Table.nftFactories, payload as any)
+    }
+    await dbClient.upsert(Table.seeds, [message])
+  } finally {
+    dbClient.close();
+  }
+}
+
+const minimumKeysPresent = (input: any, requiredKeys: any) => {
```
musnit, line 139: this logic looks like it's checking exactKeysPresent, rather than minimumKeysPresent ? that sounds fine, but if so, maybe name it exactKeysPresent?

src/utils/seedsApi.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export const persistSeed = async (payload: SeedPayload) => {
+  const dbClient = await db.init();
+  let message: any;
+  try {
+    if (payload.entity === 'platform') {
+      message = getCrdtUpdateMessage<MusicPlatform>(Table.platforms, payload as any)
+    } else if (payload.entity === 'contract') {
+      message = getCrdtUpdateMessage<NftFactory>(Table.nftFactories, payload as any)
+    }
```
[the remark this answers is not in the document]
musnit: (with validation that the table does exist)

src/utils/seedsApi.ts
```diff
@@ -0,0 +1,145 @@
...
+  const addresses = process.env.PERMITTED_ADMIN_ADDRESSES?.toLowerCase();
+  if (!addresses) {
+    throw new Error('PERMITTED_ADMIN_ADDRESSES not set');
+  }
+  return addresses.split(',');
+}
+
+
+export const validateSeed = (payload: SeedPayload) => {
+  if (payload.entity === 'platform') {
```
musnit, line 101: maybe can make these dynamic too, with higher order functions ie something like:
```suggestion

const typeValidator = (type:any, key: string) => (data: any) => {
  return type => Object.values(type).includes(data[key])
}

  const validPlatformType = typeValidator(MusicPlatformType, 'type')

  const validatorFunctions = {
  'platform': [minimumKeysPresent, validPlatformType],
  'nftFactory': [minimumKeysPresent, validContractType, validNFTStandard],
  }
  const validators = validatorFunctions[payload.entity];
  validators.forEach(validator => validator(payload.data));
  ```

File: src/utils/seedsApi.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (puffo) [reply src/utils/seedsApi.ts]: Just added the remaining parts to this. We've got `upsert` on the platforms + nftFactories, but only `update` on the artistProfiles and processedTracks. 

Will see if we get 3 or more doing the same thing, then I think it justifies making it dynamic.

File: src/utils/seedsApi.ts (hunk 8 above)
Comment (puffo) [reply src/utils/seedsApi.ts:111]: This is nice, will implement at the end as cleanup

File: src/types/seeds.ts [hunk 9]
```diff
@@ -0,0 +1,119 @@
...
+      () => typeValidator(payload, 'contractType', NFTContractTypeName),
+      () => typeValidator(payload, 'standard', NFTStandard),
+    ],
+    'artistProfiles': [
+      () => minimumKeysPresent(payload, Object.values(ArtistProfilesRequiredKeys)),
+      () => onlyValidKeysPresent(payload, ArtistProfileKeys),
+    ],
+    'processedTracks': [
+      () => minimumKeysPresent(payload, Object.values(ProcessedTracksRequiredKeys)),
+      () => onlyValidKeysPresent(payload, ProcessedTrackKeys),
```
puffo, line 70: `minimumKeysPresent` and `onlyValidKeysPresent` are used in each one of the supported seed entities. I imagine they'll get replaced with a more robust runtime TS/Schema validator library in future.

test/serve_seeds_api.ts
```diff
@@ -0,0 +1,276 @@
...
+import web3 from 'web3';
+
+import { MusicPlatformType } from '../src/types/platform';
+
+describe('Seeds API', () => {
+  let app: any;
+  const Web3 = new web3();
+
+  // TODO: rather stub out the permittedAdminAddresses() to contain this address
+  // instead of having to put the known address into .env
```
puffo, line 11: Input will be appreciated on whether there is a safe and standard way of either stubbing the .env or whether I should just go ahead and stub out the `permittedAdminAddresses` method so that we aren't dependent on `.env` configuration to make tests like this pass

test/serve_seeds_api.ts
```diff
@@ -0,0 +1,276 @@
...
+  const testAdminWallet = {
+    address: '0x8eb97c37B0BDe7A09eA5b49D6D97cd57e10559ba',
+    privateKey: '0xe07cc69757e3b261ffeb70df20f832ae74da57e11dd440a5da75377abe8caefc',
+  }
+
+  const wallet = Web3.eth.accounts.privateKeyToAccount(testAdminWallet.privateKey)
+
+  // TODO: spin up a test DB instead of using development
```
puffo, line 19: As mentioned, I could use some help with this. Alternatively, we skip it.

test/serve_seeds_api.ts
```diff
@@ -0,0 +1,276 @@
...
+        it('returns a 200', () => {
+          const body = { entity: 'platforms', data: { id: 'jamboni', name: 'Jamboni Jams', type: MusicPlatformType.sound } }
+          const signature = wallet.sign(JSON.stringify(body)).signature;
+
+          supertest(app).post(endpoint).send(body)
+            .set('x-signature', signature)
+            .expect(200)
+            .end((err,res) => { if (err) throw err });
+        })
+        it('persists the seed');
```
puffo, line 107: Related to the test db setup, we might also add an assertion to make sure that the DB persists the expected seed. But then we need to setup & flush the DB after each test. 

Alternatively, I seem to remember there are some in-memory postgres emulations for testing environments? I'll wait for some advice before doing this, so just a pending test for now.

src/utils/seedsApi.ts
```diff
@@ -126,17 +146,21 @@ export const persistSeed = async (payload: SeedPayload) => {
   let message: any;
   try {
     if (payload.entity === 'platform') {
-      message = getCrdtUpdateMessage<MusicPlatform>(Table.platforms, payload.data as any)
+      message = getCrdtUpsertMessage<MusicPlatform>(Table.platforms, payload.data as any)
     } else if (payload.entity === 'contract') {
```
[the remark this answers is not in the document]
musnit: Yeah true. We can work out if we want a translation layer on the API later on, though I do think the terms should match what they are conceptually, and conceptually these are not all contracts.

scripts/create_db.ts
```diff
@@ -5,8 +5,9 @@ import config from '../src/db/knexfile';
 import { createDB } from '../src/db/sql-db';
 
 const start = async () => {
-  const currentConfig = config[process.env.NODE_ENV]
-  console.log('Creating...');
+  const currentEnv = process.env.NODE_ENV;
+  const currentConfig = config[currentEnv]
+  console.log(`Creating ${currentEnv} environment database: '${currentConfig.connection.database}'...`);
```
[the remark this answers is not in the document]
musnit: Yeah - may be related to our other ideas around running an in-memory version of Postgres for temporary experimentation too. We can follow up on these ideas later after this PR

src/serve_seeds_api.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  try {
+    await persistSeed(req.body)
+    res.sendStatus(200);
+  } catch (e: any) {
+    res.status(500).send({ error: e.message });
+  }
+});
+
+if (process.env.NODE_ENV !== 'test') {
```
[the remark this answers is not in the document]
musnit: I'm not really familiar with supertest, but yeah agree - maybe we can find a workaround/config on it so that we don't need to add this custom test conditional

src/types/artist.ts
```diff
@@ -16,6 +16,8 @@ export type ArtistProfile = TimeField & {
   websiteUrl?: string;
 }
 
+export const ArtistProfileKeys = ['platformInternalId', 'artistId', 'name', 'platformId', 'avatarUrl', 'websiteUrl'];
```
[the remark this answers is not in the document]
musnit: Yeah i'm not sure there's a perfect runtime library that fits neatly with types - this is fine for now I think

src/types/seeds.ts
```diff
@@ -0,0 +1,119 @@
...
+enum NFTFactoriesRequiredKeys {
+  ID = 'id',
+  PLATFORM_ID = 'platformId',
+  CONTRACT_TYPE = 'contractType', // validated against `NFTContractTypeName`
+  STANDARD = 'standard', // validated against `NFTStandard`
+  AUTO_APPROVE = 'autoApprove',
+  APPROVED = 'approved'
+}
+
+enum ArtistProfilesRequiredKeys {
```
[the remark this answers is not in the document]
musnit: Agree we do need to work out a clear plan re the async nature of seeds/indexing and separation of concerns between them with validation.

**On foreign ids:**

In this case with missing foreign ids, i guess some options that come to mind are:
 - Query the DB to validate their existence as part of validation, and reject
 - Allow them through, let them fail when processed and have no effect
 - Allow them through, but update the trigger that processes messages so that it does a more general filter that filters out messages with missing foreign keys, assuming they're not ready for processing yet but that they may be available later

For this PR, I don't think we need to support artist profiles, just artists, where there shouldn't be any foreign keys _(and so then we'll probably move stuff like avatarURL and websiteURL up to artist instead of artistProfile maybe after this PR)._

**Re the update/upsert:**
Does it work to just have a rule like:
 - if a message is an upsert, all keys are required and validate them all
 - if a message is an update, at least one key is required, so just validate that

src/types/seeds.ts
musnit: ye seems fine

test/serve_seeds_api.ts
musnit: maybe we could have a seperate .env.test file and then have that override the base .env when running a test? not sure best way to do it - does something like source .env && source .env.test && ... work in the package.json ?

test/serve_seeds_api.ts
musnit: skip for now i think?

test/serve_seeds_api.ts
musnit: yeah i think let's not worry about this for now?

puffo: We can leave for now and make another pass on it if/when we do the in-memory postgres experiment

src/types/seeds.ts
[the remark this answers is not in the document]
puffo: **Foreign IDS:**
I'm thinking for the sake of closing this PR off before it grows any larger, we go for :
> Allow them through, let them fail when processed and have no effect

We can then follow up with adjusting the filter to be more selective and robust afterwards - especially seeing as using this internally for now and will likely be pretty careful with the json payloads.

**Support for artists vs artistProfiles**
We might have a similar problem with `tracks`, so we'll need to sort this out eventually. I reckon it won't be too complex to add `artists` to this PR as well. I'll give it a go and if it becomes complex, I'll bail and just reduce the scope of artistProfiles to artists.

**Re the update/upsert:**
That should work. I'll attempt a small extraction of one or two of the existing `/migrations` to confirm whether these validation options will work as we need.  They're the closest thing we have to a practical use case for now and I'm sure we'll need to extend validation as we discover more use cases to cover.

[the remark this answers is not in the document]
musnit: Cool also keep in mind that artistProfiles is kinda irrelevant once we have spinamp-centric ones - we may even be able to deprecate that entity entirely in future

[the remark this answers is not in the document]
musnit: > We might have a similar problem with `tracks`, so we'll need to sort this out eventually. I reckon it won't be too complex to add `artists` to this PR as well. I'll give it a go and if it becomes complex, I'll bail and just reduce the scope of artistProfiles to artists.

yeh for tracks i think we should only allow updates, and should not allow any foreign key fields for now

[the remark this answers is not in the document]
puffo: I think I just realised the scope of the problem changes with ArtistProfiles in that we don't have an `id` on it, so updates are a problem inherently. I've yanked it from this PR. 

Agree that we should move or deprecate fields from `ArtistProfile` as a follow up.

src/seeds/types.ts
```diff
@@ -0,0 +1,107 @@
...
+  'artists',
+  'processedTracks'
+}
+type SeedEntity = keyof typeof SeedEntities;
+
+const PlatformRequiredKeys = ['id', 'name', 'type'];
+const PlatformValidKeys = PlatformRequiredKeys;
+
+const NFTFactoryRequiredKeys = ['id', 'startingBlock', 'platformId', 'contractType', 'name', 'symbol', 'typeMetadata', 'standard', 'autoApprove', 'approved'];
+const NFTFactoryValidKeys = NFTFactoryRequiredKeys;
```
musnit, line 19: so maybe smth like;

```suggestion
const NFTFactoryRequiredKeys = ['id', 'startingBlock', 'platformId', 'contractType', 'standard', 'autoApprove', 'approved'];
const NFTFactoryValidKeys = NFTFactoryRequiredKeys.concat('typeMetadata');
```

src/seeds/types.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const ArtistRequiredKeys = ['id'];
+const ArtistValidKeys = ['id', 'name'];
+
+const ProcessedTrackRequiredKeys = ['id'];
+const ProcessedTrackValidKeys = ['id', 'title', 'slug', 'description', 'websiteUrl', 'lossyAudioURL', 'lossyAudioIPFSHash', 'lossyArtworkURL', 'lossyArtworkIPFSHash'];
```
musnit, line 25: as discussed, should we remove slug for now?

also, this makes me realize that there's also an issue if we allow lossyAudioURL/lossyAudioIPFSHash to be changed (and same for artwork):
 - ideally, we expect these both to point to the same files, just one to a http url and the other to an ipfs hash. in practice at the moment this is always the case, because we upload the files to ipfs if they aren't there, but, for exapmle if we allow the audioURL to be changed after this has occurred, it will not necessarily re-upload the new track nor generate the new ipfsHash.

Maybe we should leave these 4 fields out for now too?

ie, something like:

```suggestion
const ProcessedTrackValidKeys = ['id', 'title', 'description', 'websiteUrl'];
```

src/seeds/types.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+export const persistSeed = async (payload: SeedPayload) => {
+  let message: any;
+  let dbClient: any;
+
+  entityValidator(payload);
+
+  if (['platforms', 'nftFactories'].includes(payload.entity)) {
```
musnit, line 90: thinking on this further - matching this by type won't actually work, even for our short term needs in this PR. 

Consider:
 - in most existing migrations, we are inserting new platforms, so the getCrdtUpsertMessage is fine
 - in migration 31, we are updating existing platforms, so we do need getCrdtUpdateMessage on platforms

So we already need to support both to utilize this API for existing migrations.

I think we'll need to specify upsert vs update on the SeedPayload and support each and validate each separately (eg: for now, upsert on an nftFactory should require all keys, as it currently does, but update should only allow approved and autoApprove keys)

src/seeds/types.ts
puffo: Huh, that's a great point. Some of these edge cases seem to point to the need for triggering side-effects when changing a field. e.g. 

- when a title changes, save the old slug for backward compatibility and generate a new one
- when changing artwork/audio, ensure that uploads and pin jobs are retried

It feels like it'll be well suited to the the CRDT system, or maybe an adjacent layer to it.

But yeah, in the meantime we must be conservative with what can be adjusted, will remove those keys 👍

src/seeds/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    throw new Error('unknown seed entity');
+  }
+  if (!Object.values(CrdtOperation).includes(input.operation)) {
+    throw new Error('must specify either `upsert` or `update` operation');
+  }
+}
+
+export const persistSeed = async (payload: SeedPayload) => {
+  let dbClient: any;
+  entityValidator(payload);
```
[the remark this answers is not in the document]
puffo: Strictly speaking no as we've run it with `validateSeed`.  

But I do like the idea of independently validating the inputs for both the exported functions in case they're used outside of the API.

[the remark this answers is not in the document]
musnit: yeh same - should we do the full validation here too then not just entityValidator?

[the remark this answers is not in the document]
puffo: Yep, great solution. Adjusted and pushed up 👍

PR #110, opened 2022-09-30
Geimaj/kota

Geimaj:
Add support for candy machine NFTs.

This also adds the `typeMetadata` field to Metafactories, since candy machines are represented by MetaFactories in our schema. The `typeMetadata` will be passed down the the created NFTFactories and used in the extractor system as normal.

(Requires updating the .env with the quicknode RPC endpoint)


src/triggers/newCandyMachineNfts.ts
```diff
@@ -0,0 +1,107 @@
...
+const MAX_URI_LENGTH = 200;
+const MAX_SYMBOL_LENGTH = 10;
+const MAX_CREATOR_LEN = 32 + 1 + 1;
+const MAX_CREATOR_LIMIT = 5;
+const MAX_DATA_SIZE = 4 + MAX_NAME_LENGTH + 4 + MAX_SYMBOL_LENGTH + 4 + MAX_URI_LENGTH + 2 + 1 + 4 + MAX_CREATOR_LIMIT * MAX_CREATOR_LEN;
+const MAX_METADATA_LEN = 1 + 32 + 32 + MAX_DATA_SIZE + 1 + 1 + 9 + 172;
+const CREATOR_ARRAY_START = 1 + 32 + 32 + 4 + MAX_NAME_LENGTH + 4 + MAX_URI_LENGTH + 4 + MAX_SYMBOL_LENGTH + 2 + 1 + 4;
+
+const TOKEN_METADATA_PROGRAM = new PublicKey('metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s');
+const CANDY_MACHINE_V2_PROGRAM = new PublicKey('cndy3Z4yapfJBmL3ShUp5exZKqR3z33thTzeNMm2gRZ');
```
musnit, line 25: should we put this and the generic candy machine methods in a candy machine client instead of in the trigger? or i guess if the only place they're likely to be used is in this trigger then it's fine to leave here

src/triggers/newCandyMachineNfts.ts
```diff
@@ -0,0 +1,107 @@
...
+const MAX_NAME_LENGTH = 32;
+const MAX_URI_LENGTH = 200;
+const MAX_SYMBOL_LENGTH = 10;
+const MAX_CREATOR_LEN = 32 + 1 + 1;
+const MAX_CREATOR_LIMIT = 5;
+const MAX_DATA_SIZE = 4 + MAX_NAME_LENGTH + 4 + MAX_SYMBOL_LENGTH + 4 + MAX_URI_LENGTH + 2 + 1 + 4 + MAX_CREATOR_LIMIT * MAX_CREATOR_LEN;
+const MAX_METADATA_LEN = 1 + 32 + 32 + MAX_DATA_SIZE + 1 + 1 + 9 + 172;
+const CREATOR_ARRAY_START = 1 + 32 + 32 + 4 + MAX_NAME_LENGTH + 4 + MAX_URI_LENGTH + 4 + MAX_SYMBOL_LENGTH + 2 + 1 + 4;
+
+const TOKEN_METADATA_PROGRAM = new PublicKey('metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s');
```
musnit, line 24: is this a global program for all of solana? or specific to candy machine? or specific to kota?

Geimaj: it's the token metadata program for metaplex tokens, so not kota specific

src/processors/default/createNftFromCandyMachine.ts
```diff
@@ -0,0 +1,52 @@
...
+      const nfts = items.map(({ metadataAccount, mintTx }) => {
+        const mintAddress = metadataAccount.mintAddress.toBase58();
+
+        const details: Omit<NFT, 'owner' | 'tokenId' | 'standard'> = {
+          id: mintAddress,
+          contractAddress: mintAddress,
+          platformId: metaFactory.platformId,
+          tokenMetadataURI: metadataAccount!.uri,
+          tokenURI: metadataAccount!.uri,
+          approved: true,
```
Geimaj, line 29: approved by default

src/types/fieldExtractor.ts
```diff
@@ -161,13 +161,18 @@ const idStrategy: StrategyExtractor = (contract) => {
...
-  if (!trackId) {
+  const id = slugify(extractor(nft));
+  if (!id) {
     throw new Error(`ID not extracted correctly for nft: ${nft.id}`);
   }
-  return ethereumTrackId(nft.contractAddress, trackId);
+
+  // prefer the collection override over nft.contract address.
+  // this helps us handle dedeuplicating solana nfts (where each nft has a different contract)
+  const address = contract.typeMetadata?.collection || nft.contractAddress;
```
Geimaj, line 173: tracks will use the metaFactory address rather than the mint address in the ID to avoid duplicates

src/types/metaFactory.ts
```diff
@@ -117,5 +131,31 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
           }
         }
       })}
+  },
+  candyMachine: {
+    newContractCreatedEvent: '',
```
Geimaj, line 136: should maybe make this optional?

src/types/metaFactory.ts
```diff
@@ -117,5 +131,31 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
...
+        symbol: metadataAccount.symbol,
+        autoApprove: true, 
+        approved: true, 
+        typeMetadata: {
+          collection: metaFactory.id,
+          ...metaFactory.typeMetadata,
+          overrides: {
+            ...metaFactory.typeMetadata?.overrides,
+            artist: {
+              artistId: candyMachineArtistId(metadataAccount),
```
Geimaj, line 153: artist id extracted here because we don't store the metadata account to extract from later

src/processors/default/createNftFromCandyMachine.ts
musnit: should we use `metaFactory.autoApprove` instead ?

src/utils/identifiers.ts
```diff
@@ -1,19 +1,36 @@
...
   return ethereumArtistId(address) + suffix;
 }
 
 export const ethereumTransferId = (blockNumber: string | number, logIndex: string | number): string => {
   return `${Chain.ETHEREUM}/${blockNumber}/${logIndex}`;
 }
+
+export const solanaArtistId = (address: string): string => {
+  return `${Chain.SOLANA}/${formatAddress(address)}`;
+}
```
musnit, line 31:
```suggestion
export const solanaId = (address: string): string => {
  return `${Chain.SOLANA}/${formatAddress(address)}`;
}
```

src/utils/identifiers.ts
```diff
@@ -1,19 +1,36 @@
...
+
 export const ethereumArtistId = (address: string): string => {
   return `${Chain.ETHEREUM}/${formatAddress(address)}`;
 }
 
-export const ethereumTrackId = (address: string, trackId: string): string => {
-  const suffix = trackId !== '' ? `/${trackId}` : '';
+export const ethereumTrackId = (address: string, id: string): string => {
+  const suffix = id !== '' ? `/${id}` : '';
   return ethereumArtistId(address) + suffix;
```
musnit, line 22:
```suggestion
  return ethereumId(address) + suffix;
```

src/utils/identifiers.ts
```diff
@@ -1,19 +1,36 @@
...
+  return contract.standard === NFTStandard.METAPLEX ? solanaArtistId(address) : ethereumArtistId(address)
+}
+
+export const trackId = (contract: NftFactory, address: string, id: string): string => {
+  return contract.standard === NFTStandard.METAPLEX ? solanaTrackId(address, id) : ethereumTrackId(address, id)
+}
+
 export const ethereumArtistId = (address: string): string => {
   return `${Chain.ETHEREUM}/${formatAddress(address)}`;
 }
```
musnit, line 18:
```suggestion
export const ethereumId = (address: string): string => {
  return `${Chain.ETHEREUM}/${formatAddress(address)}`;
}
```

src/utils/identifiers.ts
```diff
@@ -1,19 +1,36 @@
 import * as slugifyLibrary from 'slugify'
 
 import { formatAddress } from '../types/address';
-import { Chain } from '../types/nft';
+import { Chain, NftFactory, NFTStandard } from '../types/nft';
 
 export const slugify = (input: string) => slugifyLibrary.default(input, { lower: true, strict: true })
 
+export const artistId = (contract: NftFactory, address: string): string => {
+  return contract.standard === NFTStandard.METAPLEX ? solanaArtistId(address) : ethereumArtistId(address)
```
musnit, line 9:
```suggestion
  return contract.standard === NFTStandard.METAPLEX ? solanaId(address) : ethereumArtistId(address)
```

src/utils/identifiers.ts
```diff
@@ -1,19 +1,36 @@
...
+}
+
+export const trackId = (contract: NftFactory, address: string, id: string): string => {
+  return contract.standard === NFTStandard.METAPLEX ? solanaTrackId(address, id) : ethereumTrackId(address, id)
```
musnit, line 13:
```suggestion
  return contract.standard === NFTStandard.METAPLEX ? solanaId(address, id) : ethereumId(address, id)
```

src/db/migrations/39-add-kota.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const up = async (knex: Knex) => {
+
+  await knex.schema.alterTable(Table.metaFactories, table => {
+    table.jsonb('typeMetadata')
+  })
+
+  await addPlatform(knex, KOTA_PLATFORM);
+  await addMetaFactory(knex, KOTA);
```
puffo, line 44: Can we switch to a CRDT instead of doing direct platform + metafactory injection? Hoping to rely entirely on seeds for new platforms and contracts and building out the parts that don't work as expected.

src/types/metaFactory.ts
```diff
@@ -117,5 +131,31 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
...
+        id: metadataAccount.mintAddress.toBase58(),
+        contractType: NFTContractTypeName.candyMachine,
+        platformId: metaFactory.platformId,
+        standard: NFTStandard.METAPLEX,
+        name: metadataAccount.name,
+        symbol: metadataAccount.symbol,
+        autoApprove: true, 
+        approved: true, 
+        typeMetadata: {
+          collection: metaFactory.id,
```
musnit, line 148: instead of collection, lets go with something like:

image

src/types/metaFactory.ts
```diff
@@ -22,18 +24,30 @@ export type MetaFactory = Contract & {
...
   autoApprove: boolean;
+  typeMetadata?: TypeMetadata;
 }
 
 export type MetaFactoryType = {
   newContractCreatedEvent: string,
   creationEventToNftFactory?: (event: ethers.Event, autoApprove: boolean, factoryMetadata?: unknown) => NftFactory
-  metadataAPI?: (events: ethers.Event[], clients: Clients) => Promise<any>
+  metadataAPI?: (events: ethers.Event[], clients: Clients) => Promise<any>,
+  metadataAccountToNftFactory?: (metadata: Metadata, metaFactory: MetaFactory) => NftFactory
```
musnit, line 34: let's try merge creationEventToNftFactory and metadataAccountToNftFactory so that they conform to the same interface.

i guess this means that instead of accepting a specific ethers.Event[] for the creationEvent it should be more generic, like `creationEventMetadataToNftFactory`

So it could look like

`creationMetadataToNftFactory?: (creationData: unknown, autoApprove: boolean, factoryMetadata?: unknown) => NftFactory `

src/types/metaFactory.ts
musnit: cool

src/triggers/newCandyMachineNfts.ts
```diff
@@ -0,0 +1,51 @@
...
+      [
+        'where', ['platformId', metaFactory.platformId]
+      ]
+    ])).map(nftFactory => nftFactory.id));
+
+
+  const allMintAccounts = await clients.solana.getMintAddressesForCandyMachine(new PublicKey(metaFactory.id))
+  const newMintAccounts = allMintAccounts.filter(mintAccount => !existingNFTFactories.has(mintAccount));
+  console.log('new candy machine nfts', newMintAccounts)
+  const chunks = _.chunk(newMintAccounts, 100)
```
puffo, line 27: Should we extract `100` to a well named ENV like `MAX_CANDY_MACHINE_CHUNK_SIZE`? 

Or alternatively use `MAX_INSERT_CHUNK_SIZE`?

src/utils/identifiers.ts
```diff
@@ -6,31 +6,31 @@ import { Chain, NftFactory, NFTStandard } from '../types/nft';
...
-  return contract.standard === NFTStandard.METAPLEX ? solanaArtistId(address) : ethereumArtistId(address)
+  return contract.standard === NFTStandard.METAPLEX ? solanaId(address) : etheruemId(address)
 }
 
 export const trackId = (contract: NftFactory, address: string, id: string): string => {
   return contract.standard === NFTStandard.METAPLEX ? solanaTrackId(address, id) : ethereumTrackId(address, id)
 }
 
-export const ethereumArtistId = (address: string): string => {
+export const etheruemId = (address: string): string => {
```
puffo, line 16:
```suggestion
export const ethereumId = (address: string): string => {
```

🙈

Geimaj: whoops! nice catch

src/types/fieldExtractor.ts
```diff
@@ -154,20 +166,32 @@ const idStrategy: StrategyExtractor = (contract) => {
...
 
-export const resolveEthereumTrackId: Resolver = (nft, contract) => {
+export const resolveTrackId: Resolver = (nft, contract) => {
+  if (isParameterizedExtractor(contract.typeMetadata?.overrides.extractor?.id)){
+    const type = (contract.typeMetadata?.overrides.extractor?.id as ParameterizedExtractorType<IdExtractorTypes>)
+    const extractorType = type.extractor;
+    const extractor = idExtractors[extractorType]({ ...type.params, nftFactory: contract }) as Extractor;
+    return extractor(nft)
+  }
+
```
musnit, line 187: think this code should go inside the idStrategy method?

(can then use the idExtractorOverride variable there and undefined/null check in there too instead of doing it twice)

also, some variable name change suggestions - maybe `ParameterizedExtractorConfig` is cleaner than `ParameterizedExtractorType`:

```suggestion
  if (isParameterizedExtractor(idExtractorOverride)){
    const config = (idExtractorOverride as ParameterizedExtractorConfig<IdExtractorTypes>)
    const extractorType = config.extractor;
    const extractor = idExtractors[extractorType]({ ...config.params, nftFactory: contract }) as Extractor;
    return extractor(nft)
  }

```

src/types/metaFactory.ts
```diff
@@ -117,5 +130,39 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
...
+            ...metaFactory.typeMetadata?.overrides,
+            extractor: {
+              id: {
+                extractor: IdExtractorTypes.USE_METAFACTORY_AND_TITLE_EXTRACTOR,
+                params: { 
+                  metaFactory: metaFactory.id,
+                  titleExtractor: metaFactory.typeMetadata?.overrides.extractor?.title
+                }
+              },
+              ...metaFactory.typeMetadata?.overrides.extractor,
```
musnit, line 157: So, the first specified `titleExtractor` sets a titleExtractor field here in the params for this parameterized extractor from the metaFactory

The second `...metaFactory.typeMetadata?.overrides.extractor` sets the usual titleExtractor override in this nftFactory also coming from the metaFactory.

The first one look unused (the implementation of the USE_METAFACTORY_AND_TITLE_EXTRACTOR is pulling the titleStrategy from the second set field, not the first)

So, todos:
 - Decide which one we think is better? I think the second one which is actually being used is maybe better?
 - Remove the first one, since it's unused?

src/types/metaFactory.ts
```diff
@@ -117,5 +130,39 @@ export const MetaFactoryTypes: MetaFactoryTypes = {
...
+        approved: autoApprove, 
+        typeMetadata: {
+          ...metaFactory.typeMetadata,
+          overrides: {
+            ...metaFactory.typeMetadata?.overrides,
+            extractor: {
+              id: {
+                extractor: IdExtractorTypes.USE_METAFACTORY_AND_TITLE_EXTRACTOR,
+                params: { 
+                  metaFactory: metaFactory.id,
```
musnit, line 153:
```suggestion
                  metaFactoryId: metaFactory.id,
```

src/types/fieldExtractor.ts
```diff
@@ -79,6 +87,10 @@ const idExtractors: IdExtractorMapping = {
     }
     return `${nft.metadata.trackNumber}`
   },
+  [IdExtractorTypes.USE_METAFACTORY_AND_TITLE_EXTRACTOR]: (params: { metaFactory: string, nftFactory: NftFactory }) => (nft: NFT) => {
+    const title = titleStrategy(params.nftFactory)(nft);
+    return solanaTrackId(params.metaFactory, title);
+  }
 }
```
musnit, line 94:
```suggestion
  [IdExtractorTypes.USE_METAFACTORY_AND_TITLE_EXTRACTOR]: (params: { metaFactoryId: string, nftFactory: NftFactory }) => (nft: NFT) => {
    const title = titleStrategy(params.nftFactory)(nft);
    return solanaTrackId(params.metaFactoryId, title);
  }
}
```

src/types/fieldExtractor.ts
musnit: ah right, cool. i guess the confusion is that this function resolveTrackId is basically the trackIdStrategy and idStrategy is actually more like trackIdSuffixStrategy - maybe should rename them as such?

PR #117, opened 2022-10-24
Fix Sound GQL query

puffo:


musnit:
@puffo is this code still used?

puffo:
The clean rebuild was failing to fetch metadata due to a  GQL validation error. Do you think we should do away with it entirely?

musnit:
> The clean rebuild was failing to fetch metadata due to a GQL validation error. Do you think we should do away with it entirely?

was it from this function or somewhere else?

and yeah, potentially migrating to the multi-print that does no API query for the legacy sound.xyz tracks could be the best solution, though i haven't dug into it yet

puffo:
Ok, this is not as much of a one liner as I'd hoped. Closing the PR for now and we'll tackle sound deprecations as a separate issue.

https://github.com/spinamp/spindexer/issues/118

PR #120, opened 2022-10-28
Configure CORS for POST

puffo:


src/seeds/server.ts
```diff
@@ -10,8 +10,12 @@ const apiVersionPrefix = `/v${process.env.SEEDS_API_VERSION || '1'}`;
 
 export const createSeedsAPIServer = () => {
   const app = express();
+  app.use(cors({
+    'origin': true,
```
musnit, line 14: should be false?

PR #121, opened 2022-10-29
Persist signer address when adding a seed message

puffo:


src/constants/defaults.ts
```diff
@@ -0,0 +1,5 @@
+import { ETHEREUM_NULL_ADDRESS } from '../types/ethereum'
+
+export const defaultAdminAddress = () => {
+  return process.env.DEFAULT_ADMIN_ADDRESS?.toLowerCase() || ETHEREUM_NULL_ADDRESS
+}
```
musnit, line 5: don't need to use a function here, can just use the value directly?

also, should maybe just throw error & crash if there is no DEFAULT_ADMIN_ADDRESS set?

src/db/migrations/31-zora-approvals.ts
```diff
@@ -28,12 +27,13 @@ const pmRecords = [
   '0x14214694847ba90da2e9b08b6de397bd06d5a626',
   '0x676777d784ebcc9d3ff3e28ce73757864a8333b4']
 
-
 export const up = async (knex: Knex) => {
-
-  const messages: CrdtMessage[] = pmRecords.map(address =>
-    getCrdtUpdateMessage<NftFactory>(Table.nftFactories, { id: address, approved: true, autoApprove: true })
-  )
+  const messages: any[] = pmRecords.map(address => ({
```
musnit, line 31: can just use the defaultAdminAddress for these so that we can keep using getCrdtUpdateMessage instead of manually doing it?

puffo: We've got the same recurring problem when retrofitting migrations here - the `signer` column doesn't yet exist so if we modify the `getCrdtUpdateMessage` signature, we break previous usages in our migrations. Alternatively we can shift the order of migrations to change migration history but that's also bad practice.

Ideal scenario for migrations is that they deal purely with data, so I think this is a positive adjustment towards decoupling our migrations from our general business logic and maintaining a more chronological migration path.

src/seeds/middleware.ts
```diff
@@ -1,25 +1,25 @@
...
-    let signer: any;
-
     const auth = {
       message: JSON.stringify(request.body),
       signature: request.header('x-signature') || '',
     }
 
     try {
-      signer = validateSignature(auth, permittedAdminAddresses());
+      request.signer = validateSignature(auth, permittedAdminAddresses());
```
musnit, line 22: In the next PR, can we go ahead and add an API endpoint for the zora approvals so that we can make it public? (I imagine a new route specifically for approving contracts that allows for anyone to submit).

There's a few different ways we can do this, so plz think about it and let me know if you have a better suggestion, but here's one suggested design:
 - Message-based authorization, rather than global authorization
   - validateSignature just recovers and verifies the signature and puts it on the request
   - move middleware.ts#L50 into it's own function called onlyAdmin
   - Rename /seeds/ to /messages/ so that it's a bit more generic
   - Rename validateSeed  -> validateMessage
   - POST /messages/ endpoint does the onlyAdmin check within the validateMessage as part of validation
  - New message type for approving Zora Contracts
    - There is a 1:1 mapping from SeedPayload.CrdtOperation to crdtOperationMessageFnMap, we want to separate this since the Zora approvals will use a new payload operation, but will still be persisted to the DB as a basic update.
    - Create a new message type called CONTRACT_APPROVAL
    - For this operation, the payload should only allow nftFactories and should require exactly NFTFactoryValidUpdateKeys in the payload and should verify that the provided contract is a Zora contract. It can also skip the admin check.

musnit:
2 small comments, and one for the next PR, but otherwise looks good!

src/db/migrations/31-zora-approvals.ts
musnit: oh right, sorry, sounds fine then

src/seeds/middleware.ts
puffo: Already have the next one in the pipeline! https://github.com/spinamp/spindexer/pull/122

I agree with the new message type for CONTRACT_APPROVAL, and having a per-message authorization that isn't restricted to the admins.

PR #123, opened 2022-11-01
update query 🐞

Geimaj:
remove `signer` field from newSeedMessages query

### DIFF
--- a/src/db/migrations/41-add-signer-to-mempool.ts
+++ b/src/db/migrations/41-add-signer-to-mempool.ts
@@ -1,0 +1,21 @@
+import { Knex } from 'knex';
+
+import { Table } from '../db';
+import { tableNameToViewName, updateViews } from '../migration-helpers';
+
+export const up = async (knex: Knex) => {
+  await knex.schema.alterTable(Table.mempool, table => {
+    table.string('signer');
+  })
+  await knex(Table.mempool).update({ signer: process.env.DEFAULT_ADMIN_ADDRESS })
+  await updateViews(knex);
+}
+
+export const down = async (knex: Knex) => {
+  await knex.schema.dropViewIfExists(tableNameToViewName(Table.mempool));
+  await knex.schema.alterTable(Table.mempool, table => {
+    table.dropColumn('signer');
+  });
+
+  await updateViews(knex);
+}
--- a/src/triggers/messages.ts
+++ b/src/triggers/messages.ts
@@ -84,7 +84,7 @@ export const pendingMempoolUpdateMessages: (tables: string) => Trigger<undefined
   }
 
 export const newSeedMessages: Trigger<Cursor> = async (clients, cursor: string) => {
-  const selectSql = `select id, "timestamp" , "table", "data" , operation from ${Table.seeds}
+  const selectSql = `select m.* from ${Table.seeds} m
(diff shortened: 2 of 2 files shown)


puffo (approved):

musnit:
or should we just put the signer field into the mempool? could be useful in future

puffo:
> @Geimaj @puffo or should we just put the signer field into the mempool? could be useful in future

We should be able to derive the mempool contents from the seed though? What use case are you thinking might be helped by having signer on the mempool too?

musnit:
> > @Geimaj @puffo or should we just put the signer field into the mempool? could be useful in future
> 
> We should be able to derive the mempool contents from the seed though? What use case are you thinking might be helped by having signer on the mempool too?

One example could be if we migrate playlists/likes into Spindexer, then we'd want the signer as part of them as it would be needed during processing. Additionally, thinking about it more, the original intention behind Seeds is that it would be used as basically a dumb-backup for records just used for backup and restore and that most messages would flow directly into the mempool rather than via Seeds _(we may wanna actually consider fixing this too, but no big deal for now)_. This would especially be the case for records that we may not want to persist indefinitely, like listening activity if we migrate that into the Spindexer in future too.

Having both seeds + mempool does mix things up a little. I think it works for now, but I imagine we may consolidate things differently in future and so keeping all fields the same so that it continues to be more like a backup/restore system I think is likely best

musnit:
> Having both seeds + mempool does mix things up a little. I think it works for now, but I imagine we may consolidate things differently in future and so keeping all fields the same so that it continues to be more like a backup/restore system I think is likely best

I think the real distinction is between functionality that should be persisted indefinitely and become part of Spindexer's core itself, vs functionality that are part of Spindexer's state

We've migrated most new platform processing functionality out of Typescript code and into database state via the new multi-track processor DSL, and seeds records, but really I still consider it to be part of Spindexer's core functionality.

For future ideas like artist profile updates, or for a favs/playlist/listening activity integration, that kind of content does feel more analogous to real stateful application content that is in a stateful DB, and so potentially wouldn't be part of seeds and only be part of mempool that is processed once and only backed-up post-processing rather than backed up pre-processing as the Seed table does.

puffo:
Those are some good points. What it comes down to for me is that we get a lot of agility in what we want to include as part of the responsibility of Spindexer by just adding the simple `signer` column to the `mempool` table.

My opinion is that spindexer should be kept as pure as possible in functionality related to aggregating blockchain and related data sources, so while I don't necessarily agree that this includes wallet-specific state (playlists etc) generated by API consumers, now isn't the right time to debate that 😄 

TLDR for me; the benefit of the flexibility far outweighs the simple cost of reverting the column if we don't want it in future 😄 

I've added a migration and tested locally against a `yarn db-reset` + `yarn start`. All good ✅

musnit:
> My opinion is that spindexer should be kept as pure as possible in functionality related to aggregating blockchain and related data sources, so while I don't necessarily agree that this includes wallet-specific state (playlists etc) generated by API consumers, now isn't the right time to debate that 😄

yeah likely agree too - but i think the latter data will also be able to use much of the same codebase and patterns with CRDTs and all that, including the mempool pattern

puffo:
All good to go on this one! Merging after 👍 via DMs

PR #125, opened 2022-11-02
turn off autoApprove

Geimaj:
disable autoApprove to avoid indexing nfts that aren't music / haven't got complete metadata


musnit:
done in https://github.com/spinamp/spindexer/pull/126

PR #127, opened 2022-11-07
Geimaj/lens

Geimaj:
DRAFT

Adds support for any EVM chain, and includes a LENS integration on polygon.

TODO: 

- [ ] change nftFactory ids to have chain prefix
- [ ] change metaFactory ids to have chain prefix
- [ ] add `address` filed to factories
- [ ] change nft ids to have chain prefix
- [ ] change chain prefix in erc721Transfers
- [ ] rename `createdAtEthereumBlockNumber` -> `createdAtBlockNumber` in transfers and nfts
- [ ] rename all instances of `ethereum` to `evm`
- [ ] rename block query gap
- [ ] use lens profile pic in artist profile
- [ ] replace graphql query with contracts query (needs discussion with lens)
- [ ] create blocks table to cache timestamp requests

Notes:
I think collector.id, artist.id doesn't need a chain prefix because the same address can be used across any evm chain


src/types/metaFactory-types/lens.ts
```diff
@@ -0,0 +1,104 @@
...
+  metadataAPI: async (events, clients, metaFactory) => {
+    const profileIds = Array.from(new Set(events.map(event => ethers.BigNumber.from(event.args!.profileId).toHexString())));
+
+    if (profileIds.length === 0){
+      return {}
+    }
+
+    const chunks = _.chunk(profileIds, 50);
+
+    const queries = chunks.map(chunk => {
```
musnit, line 24: as discussed, maybe just query the chain's ProfileCreated event here instead

src/types/metaFactory-types/lens.ts
```diff
@@ -0,0 +1,104 @@
...
+      typeMetadata: {
+        overrides: {
+          extractor: {
+            id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+            title: TitleExtractorTypes.METADATA_NAME,
+            artistName: ArtistNameExtractorTypes.USE_ARTIST_NAME_OVERRIDE,
+            artistId: ArtistIdExtractorTypes.USE_ARTIST_ID_OVERRIDE,
+          },
+          artist: {
+            name: apiMetadata.profile.handle,
```
musnit, line 95: are artist names usually not in the nft metadata too? though i guess handling via here works too

src/types/metaFactory-types/lens.ts
```diff
@@ -0,0 +1,104 @@
...
+            artistId: apiMetadata.profile.ownedBy
```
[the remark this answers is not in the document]
Geimaj: yeah correct here `ownedBy` is the EOA. Will be replaced when this changes to querying the `ProfileCreated` event

src/clients/noizd.ts
```diff
@@ -21,7 +21,7 @@ export type NOIZDClient = {
 }
 
 const mapAPIListingToTrackID = (listing: any): string => {
-  return ethereumTrackId(listing.nft.metadata.contract, listing.id);
+  return `noizd/${listing.id}`
 };
```
musnit, line 25: hmm what was going on here? was it like ethereum/uuid and so kinda broken before already? will this result in changing track ids tho?

src/types/metaFactory-types/lens.ts
Geimaj: Not consistently, but sometimes it's in the track title. We could split it out but I thought using the lens handle was a better option here

src/clients/noizd.ts
Geimaj: I see that some noizd tracks are prefixed with noizd/ and others are prefixed with ethereum/. Not exactly sure if always going with noizd/ is a good solution but not all noizd nfts are on ethereum or on chain so I think it's better to use the hardcoded noizd/ prefix here?

.env.example
```diff
@@ -15,6 +15,12 @@ GLOBAL_STARTING_TIMESTAMP=1555033358
 ETHEREUM_BLOCK_QUERY_GAP=500000
 BLOCKS_SUBGRAPH_ENDPOINT=https://api.thegraph.com/subgraphs/name/snowfork/ethereum-blocks
 
+# Polygon-specific config
+ETHEREUM_PROVIDER_ENDPOINT=https://polygon-mainnet.alchemyapi.io/v2/key
```
musnit, line 19:
```suggestion
POLYGON_PROVIDER_ENDPOINT=https://polygon-mainnet.alchemyapi.io/v2/key
```

src/clients/noizd.ts
musnit: yeah basically all tracks start with the noizd/ prefix premint but then once the nft is minted, the id gets migrated to ethereum/ and changes

(this is actually problematic too, since ids change and tracks will get lost from favs/playlists, but i wouldn't change it from how it currently is since there may be side effects in other processors)

puffo: That's a pretty surprising change in ids. 
Would it remove the need for that transformation from `noizd`->`ethereum` if we just added the platform as part of the track id, along with the chain and listing id?

i.e.
```
`ethereum/noizd/${listing.id}`
```

Or does noizd publish to different chains, making the `ethereum` part inaccurate?

puffo: But maybe the ideal form of an immutable ID here is to just have the `chain/address`, especially seeing as platform information isn't strictly part of all contracts.

Like you said though @musnit, maybe we deal with this problem later but it'll be useful to know what we're aiming for with IDs in general (due to similar considerations in the Artist and ArtistProfile part of the codebase).

puffo (commented):
Really impressive changes to support all EVMs @Geimaj ! 

I've mostly reviewed to keep up to date with our direction here, but the problem of immutable IDs definitely applies to my work as well. Keen to hear your thoughts there.

src/db/migrations/43-add-lens.ts
```diff
@@ -0,0 +1,182 @@
...
+    table.string('chainId')
+    table.string('address')
+  })
+  
+  await knex.schema.alterTable(Table.metaFactories, table => {
+    table.string('chainId')
+    table.string('address')
+  })
+
+  // update factories
```
musnit, line 75: do we also need an update to update existing nftFactories.address and metaFactories.address? maybe can do in follow up PR if needed tho

Geimaj: This migration already updates nftFactories.address and metaFactories.address:

```const nftFactoryUpdates = "
    update ${Table.nftFactories}
    set address=id,
    "chainId"= 
    case
      when "standard" = '${NFTStandard.METAPLEX}' then 'solana'
      else 'ethereum'
    end
  "
  const metaFactoryUpdates = "
    update ${Table.metaFactories}
    set address=id,
    "chainId"= 
    case
      when "standard" = '${NFTStandard.METAPLEX}' then 'solana'
      else 'ethereum'
    end
  "
```

src/runner.ts
```diff
@@ -1,28 +1,37 @@
...
   const catalogClient = await catalog.init();
   const soundClient = await sound.init();
   const noizdClient = await noizd.init();
   const solanaClient = await solana.init();
+  const dbClient = existingDBClient || await db.init();
+  
+  const evmChains = await dbClient.getRecords<Chain>(Table.chains, [['where', ['type', ChainType.evm]]])
+  const evmClients = {} as { [chainId in ChainId]: EVMClient }
+  
+  for await (const chain of evmChains){
```
musnit, line 26: oh haven't seen `for await` before, seems nice

src/runner.ts
```diff
@@ -1,28 +1,37 @@
...
 import ipfs from './clients/ipfs';
 import noizd from './clients/noizd';
 import solana from './clients/solana';
 import sound from './clients/sound';
 import { DBClient, Table } from './db/db';
 import db from './db/sql-db';
+import { Chain, ChainId, ChainType } from './types/chain';
 import { Clients, Processor } from './types/processor';
 
 export const initClients = async (existingDBClient?: DBClient) => {
```
musnit, line 14: is existingDBClient used?

src/utils/identifiers.ts
```diff
@@ -1,33 +1,35 @@
...
-export const ethereumTrackId = (address: string, id: string): string => {
+export const evmTrackId = (chainId: ChainId, address: string, id: string): string => {
   const suffix = id !== '' ? `/${id}` : '';
-  return ethereumId(address) + suffix;
+  return evmId(chainId, address) + suffix;
 }
 
-export const ethereumTransferId = (blockNumber: string | number, logIndex: string | number): string => {
-  return `${Chain.ETHEREUM}/${blockNumber}/${logIndex}`;
+export const transferId = (chainId: ChainId, blockNumber: string | number, logIndex: string | number): string => {
```
musnit, line 27: do we need to migrate existing transfer ids to prefix with ethereum?

src/runner.ts
Geimaj: yeah this is quite nice to use because array.forEach doesn't work the same way for async functions

src/runner.ts
Geimaj: yes, it's used here https://github.com/spinamp/spindexer/blob/7b19f9c634263bf994a5754976b5bfc8bd76f878/src/runner.ts#L45

src/utils/identifiers.ts
```diff
@@ -1,33 +1,35 @@
...
+  return `${chainId}/${blockNumber}/${logIndex}`;
 }
 
 export const solanaId = (address: string): string => {
```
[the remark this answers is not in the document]
Geimaj: We have always stored solana addresses as mixed case, so I don't think that will be a problem. However now we include the solana/ prefix for artistIds, so I should either remove that or update the existing data in a migration.

I imagine updating it will lead to lost follows though so probably better to remove that prefix for solana artists

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
```diff
@@ -101,18 +105,23 @@ const processorFunction = (contracts: NftFactory[]) =>
     const transfersForExistingNfts = transfers.filter(transfer => existingNfts.has(transfer.nftId!));
 
     await clients.db.insert(Table.erc721Transfers, transfersForExistingNfts);
-    await clients.db.updateProcessor(NAME, newCursor);
+    await clients.db.updateProcessor(name, newCursor);
   };
+  
+export const createERC721NFTsFromTransfersProcessor: (chainId: ChainId, contracts: NftFactory[]) => Processor = 
+  (chainId, contracts) => {
+    const cursorName = chainId === ChainId.ethereum ? NAME : `${NAME}_${chainId}`;
```
musnit, line 113: cool this is for legacy support ye?

i think we should keep a running list of legacy functionality/state to be fixed/migrated in future?

Initial list that comes to mind is:
 - Ethereum processor name in createERC721NFTsFromTransfersProcessor.ts
 - Ethereum prefixed artist ids
 - Lowercase instead of case-sensitive ethereum addresses
 - Lowercase solana addresses

src/types/platforms-types/catalog.ts
```diff
@@ -1,12 +1,13 @@
...
+import { evmTrackId, evmId, slugify } from '../../utils/identifiers';
 import { ArtistProfile } from '../artist';
+import { ChainId } from '../chain';
 import { MapNFTsToTrackIds, MapTrack } from '../mapping';
 import { NFT } from '../nft';
 
-const mapAPITrackToArtistID = (apiTrack: any): string => {
-  return ethereumId(apiTrack.artist.id);
+const mapAPITrackToArtistID = (chainId: ChainId, apiTrack: any): string => {
+  return evmId(chainId, apiTrack.artist.id);
```
musnit, line 10: should this be hardcoded to ethereum, since it's an artist ID? and use artistId() instead of evmId()?

src/types/platforms-types/sound.ts
```diff
@@ -1,13 +1,14 @@
...
+import { evmId, slugify } from '../../utils/identifiers';
 import { ArtistProfile } from '../artist';
+import { ChainId } from '../chain';
 import { MapTrack, MapNFTsToTrackIds } from '../mapping';
 import { NFT, getNFTMetadataField } from '../nft';
 
-const mapAPITrackToArtistID = (apiTrack: any): string => {
-  return ethereumId(apiTrack.artist.user.publicAddress);
+const mapAPITrackToArtistID = (chainId: ChainId, apiTrack: any): string => {
+  return evmId(chainId, apiTrack.artist.user.publicAddress);
```
musnit, line 11: should this be hardcoded to ethereum, since it's an artist ID? and done so using artistId() instead of evmId()?

src/processors/default/createNftFromCandyMachine.ts
```diff
@@ -29,6 +30,8 @@ export const createNftsFromCandyMachine: (metaFactory: MetaFactory) => Processor
           approved: metaFactory.autoApprove,
           burned: false,
           createdAtTime: new Date(mintTx.blockTime! * 1000),
+          chainId: metaFactory.chainId,
+          nftFactoryId: getFactoryId(metaFactory.chainId, mintAddress)
```
musnit, line 34: will we need to backfill this nftFactoryId field on old nfts?

src/processors/default/createNinaNftProcesor.ts
```diff
@@ -28,13 +29,17 @@ export const createNinaNfts: Processor = {
         platformId: MusicPlatformType.nina,
         tokenMetadataURI: metadataAccount.uri,
         tokenURI: metadataAccount.uri,
-        approved: true // all nina nfts approved by default
+        approved: true, // all nina nfts approved by default
+        chainId: ChainId.solana,
+        nftFactoryId: mintAddress
```
musnit, line 34: will we need to backfill this nftFactoryId field on old nfts?

src/processors/default/createNftFromCandyMachine.ts
Geimaj: yes we do this in the lens migration

src/processors/default/createNinaNftProcesor.ts
Geimaj: we backfill all nftFactoryIds in the lens migration

src/processors/default/processPlatformTracks/processPlatformTracks.ts
```diff
@@ -18,9 +18,9 @@ const getNFTFactories = async (nfts: NFT[], dbClient: DBClient) => {
   if (nfts.length === 0) {
     throw new Error('Unexpected empty NFT array');
   }
-  const nftFactoryAddresses = _.uniq(nfts.map((n: NFT) => n.contractAddress));
-  const nftFactoryAddressesString = JSON.stringify(nftFactoryAddresses).replace(/\"/g, "'").replace('[', '(').replace(']', ')');
-  const query = `select * from "${Table.nftFactories}" where id in ${nftFactoryAddressesString}`
+  const nftFactoryIds = _.uniq(nfts.map((n: NFT) => n.nftFactoryId));
```
musnit, line 21: do we migrate all  n.contractAddress =>  n.nftFactoryId properly to ensure they all are as expected?

Geimaj: yes I did this in the migration. We rename contractAddress -> nftFactoryId, and then we update contractAddress to equal nftFactoryId. So everything will be updated as we expect

src/types/metaFactory-types/lens.ts
```diff
@@ -0,0 +1,102 @@
...
+      platformId: metaFactory.platformId,
+      standard: metaFactory.standard,
+      startingBlock: `${parseInt(event.blockNumber) - 1}`,
+      typeMetadata: {
+        overrides: {
+          extractor: {
+            id: IdExtractorTypes.USE_TITLE_EXTRACTOR,
+            title: TitleExtractorTypes.METADATA_NAME,
+            artistName: ArtistNameExtractorTypes.USE_ARTIST_NAME_OVERRIDE,
+            artistId: ArtistIdExtractorTypes.USE_ARTIST_ID_OVERRIDE,
```
musnit, line 90: do we still need these extractors given that we're overriding them below?

Geimaj: Yes because the multi-track-multi-print mappers use these resolvers, so we have to use these values to specify that the resolvers should use the override value

puffo (approved):
Looks good to me! 1 small fix needed for throwing errors in `evm.ts` and otherwise just some questions/comments!

I'm keen to get this into the main branch fairly soon to deal with merge conflicts on my side.

src/utils/identifiers.ts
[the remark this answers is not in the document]
musnit: @puffo 
1 - we'll keep the legacy format for now and not do any migration. if we do want a migration in future, it will need to happen full stack everywhere, ie, spindexer + backend + frontend cache
2 - case-sensitive and prefix are just 2 scenarios, but i think there will be more in future that can't be easily handled (eg: merging artist profiles). i think getting ourselves setup to be able to do (1) is the best long term solution

src/db/orm.ts
```diff
@@ -16,7 +17,8 @@ const toDBRecord = <RecordType>(record: RecordType | RecordUpdate<unknown>) => {
 const toRecordMapper: any = {
   [Table.nftFactories]: (erc721Contracts: NftFactory[]): IdField[] => erc721Contracts.map((c) => {
     return ({
-      id: c.standard === NFTStandard.METAPLEX ? c.id : c.id.toLowerCase(), // solana addresses are base58 encoded so are case sensitive
+      id: c.chainId === ChainId.solana ? c.id : c.id.toLowerCase(), // solana addresses are base58 encoded so are case sensitive
```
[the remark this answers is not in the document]
musnit: yeah, migrating to case-sensitive is def the eventual plan. i think case-insensitivity is just due to legacy infrastructure we used before but can be done now

musnit (approved):

src/processors/default/createERC721NFTsFromTransfersProcessor.ts
Geimaj: yeah I think some changes will be required on the FE:

- nftFactory and metaFactory now have an id field, and an address field, so to get the address use the address field not the id field
- nft.contractAddress is now the field to get the contract address on an nft
- nft.nftFactoryId is now the nftFactory id, not the contract address

I don't think the FE uses blocks for anything but all those fields were renamed to createdAtBlockNumber

musnit: cool i don't think the frontend uses any of those fields atm? @emzet93 ?

musnit: maybe nft.nftFactoryId was used for the etherscan link, so now should need to be contractAddress?

PR #128, opened 2022-11-08
Reset IPFS uploads when artwork has changed

puffo:
To improve data integrity and the ability to change the artworkURL on a track (possibly via the CRDT seeds API), this PR implements a new trigger + processor combination.

- The `ipfsArtworkResetter` detects tracks where the `lossyArtworkIPFSHash` and `lossyArtworkURL` are out of sync (with the IPFSFiles table's `cid` and `url` respectively) and clears the track's `lossyArtworkIPFSHash` 
- The cleared `lossyArtworkIPFSHash` columns are then reprocessed by the `ipfsArtworkUploader` processor and brought into parity with the IPFSFiles table

A Migration is included to backfill the existing IPFSFiles table, along with additions to the test suite  to verify processor + trigger behaviors

Outstanding TODOS:

- [ ] Adjust the processor that is currently skipping IPFS File additions so that IPFS File table is kept in parity with the referenced urls and cids for all artworks
- [ ] Decide on whether to apply this to the `lossyAudioURL` + `lossyAudioIPFSHash` for consistency

### DIFF
--- a/src/triggers/ipfs.ts
+++ b/src/triggers/ipfs.ts
@@ -60,3 +60,19 @@ export const artworkNotOnIpfs: Trigger<undefined> = async (clients: Clients) => 
 
   return tracksWithFiles
 };
+
+export const artworkChanged: Trigger<undefined> = async (clients: Clients) => {
+  const query = `select * from "${Table.processedTracks}" as t
+      inner join "${Table.ipfsFiles}" i
+      on t."lossyArtworkIPFSHash" = i.cid
+      where t."lossyArtworkIPFSHash" is not null
+      and t."lossyArtworkURL" is not null
+      and t."lossyArtworkURL" <> i.url
+      LIMIT ${process.env.IPFS_UPLOAD_BATCH_SIZE || process.env.QUERY_TRIGGER_BATCH_SIZE!}`
+
+  const tracksWithChanges = (await clients.db.rawSQL(
+    query
+  )).rows
+
+  return tracksWithChanges
+};
--- a/src/triggers/ipfs.ts
+++ b/src/triggers/ipfs.ts
@@ -63,11 +63,13 @@ export const artworkNotOnIpfs: Trigger<undefined> = async (clients: Clients) => 
 
 export const artworkChanged: Trigger<undefined> = async (clients: Clients) => {
   const query = `select * from "${Table.processedTracks}" as t
-      inner join "${Table.ipfsFiles}" i
-      on t."lossyArtworkIPFSHash" = i.cid
-      where t."lossyArtworkIPFSHash" is not null
-      and t."lossyArtworkURL" is not null
-      and t."lossyArtworkURL" <> i.url
+      left outer join "${Table.ipfsFiles}" i
+      on t."lossyArtworkURL" = i.url
+      where (
+        (t."lossyArtworkIPFSHash" is not null and t."lossyArtworkURL" is not null)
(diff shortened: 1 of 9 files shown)


src/triggers/ipfs.ts
```diff
@@ -60,3 +60,21 @@ export const artworkNotOnIpfs: Trigger<undefined> = async (clients: Clients) =>
 
   return tracksWithFiles
 };
+
+export const artworkChanged: Trigger<undefined> = async (clients: Clients) => {
```
musnit, line 64: to clarify:
 - this trigger essentially checks for any tracks where the track has both a lossyArtworkIPFSHash and a lossyArtworkURL, but the ipfs file is missing in the ipfsFiles table?

Some concerns:
 a - Naming wise, rather than calling artworkChanged I'd call it artworkMismatch or something, since it's reflecting a current state of the database rather than a change or an event.
 b - When used together with a feature where a track URL is changed, this also means that the database will be in this temporary/intermediary state before this trigger runs where there is a mismatch with bad data. Doesn't seem ideal and makes me think we should handle this logic at the time of update rather than later, though that may have its own complexities.
 c - In practice, I'm not sure this will actually work. The ipfsFiles table only has records for files we upload ourselves that are not already on IPFS. There are many examples of tracks which will have both a lossyArtworkIPFSHash and lossyArtworkURL (eg: Catalog tracks are always already on IPFS, so instead of uploading them, we just set the lossyArtworkIPFSHash directly in another trigger/processor), and so I think they will get caught by this trigger when they shouldn't? Also, I think it may be the case that they'll get reset by this trigger but then fixed by that other trigger without ever actually doing an insert into the db and may loop infinitely?

I'd say lets think more about a solution where we handle this logic at the time of Update and if there's an easy way to do that, then maybe go with that rather?

puffo: > this trigger essentially checks for any tracks where the track has both a lossyArtworkIPFSHash and a lossyArtworkURL, but the ipfs file is missing in the ipfsFiles table?

Correct! 

> a - Naming wise ...

Cool, makes sense.

> b - ... this also means that the database will be in this temporary/intermediary state

I don't think this a problem we can avoid given our dependencies on different external data source, and in fact it already happens regularly. When an API fetch fails and we then populate an `error` field (only to attempt to clear and reprocess later), it is effectively the same as this bad/mismatched data problem, no? 

I recognise that the cause of these inconsistencies is different (i.e. API being down, vs an incoming change from a CRDT message), but the proposed mechanism is in line with our existing self-healing and robustness processors.

> c - ... The ipfsFiles table only has records for files we upload ourselves that are not already on IPFS

Right, but at a conceptual level @Geimaj and I discussed whether the ipfsFiles table should contain _only_ the stuff we've uploaded ourselves, or become a _general_ source of truth for all ipfsFiles associated to entities in spinamp (tracks, artists, etc). 

We decided on the latter, and as such included a migration to backfill existing tracks (`src/db/migrations/43-backfill-ipfs-table.ts`), and to adjust the direct assignment in that trigger/processor you mentioned [still TODO].

puffo: I can definitely put together a POC to adjust this at the API level for comparison. 

I had thought about that as a possible solution but ultimately my main concern there was that the business logic responsible for the validation of our internal spindexer state was now leaking across to the Seeds API instead of being self contained in the processor queue.

PR #129, opened 2022-11-11
Customizable Artist fields

puffo:


src/types/artist.ts
```diff
@@ -4,6 +4,11 @@ import { Record, TimeField } from './record'
 
 export type Artist = Record & {
   name: string;
+  address?: string;
+  avatarUrl?: string;
+  externalLinks: any[]; // TODO: build out a TS type with name: string, type: string, link: string
+  theme: any[]; // TODO: check for an existing TS type or define one
```
musnit, line 10: i think only in the frontend codebase

src/types/artist.ts
```diff
@@ -4,6 +4,11 @@ import { Record, TimeField } from './record'
...
+  spinampLayoutConfig: any[]; // TODO build out a TS type for type: string, data: string, visible: boolean
```
musnit, line 11: maybe let's leave it dynamic for now as just any (not any[]) since i think it may change a bunch in time?

src/types/artist.ts
```diff
@@ -4,6 +4,11 @@ import { Record, TimeField } from './record'
 
 export type Artist = Record & {
   name: string;
+  address?: string;
+  avatarUrl?: string;
+  externalLinks: any[]; // TODO: build out a TS type with name: string, type: string, link: string
```
musnit, line 9: can be optional?

src/types/artist.ts
musnit: also can be optional?

src/types/artist.ts
musnit: also can be optional?

src/types/artist.ts
```diff
@@ -16,11 +21,13 @@ export type ArtistProfile = TimeField & {
   websiteUrl?: string;
 }
 
-export const mapArtist = (artistProfile: ArtistProfile): Artist => {
+export const mapArtist = (artistProfile: ArtistProfile): Partial<Artist> => {
```
musnit, line 24: hmm should this be Partial? shouldn't we expect a full/complete artist object always?

src/utils/identifiers.ts
```diff
@@ -33,4 +34,21 @@ export const solanaId = (address: string): string => {
...
+    return id
+  }
+
+  const parts = id.split('/');
+  if (parts[1] && isControlledEthereumAddress(parts[1])) {
+    return parts[1];
+  }
+}
+
+const isControlledEthereumAddress = (address: string): boolean => {
```
musnit, line 52: can we just use https://docs.ethers.io/v5/api/utils/address/#utils-isAddress ?

src/utils/identifiers.ts
```diff
@@ -33,4 +34,21 @@ export const solanaId = (address: string): string => {
 export const solanaTrackId = (address: string, id: string): string => {
   const suffix = id !== '' ? `/${id}` : '';
   return solanaId(address) + suffix;
-}
\ No newline at end of file
+}
+
+export const controlledEthereumAddressFromId = (id: string | undefined): string | undefined => {
```
musnit, line 39: i'd just call this ethereumAddressFromId since there's nothing control-specific about it

musnit: also this should be a string, never undefined

src/types/artist.ts
```diff
@@ -18,10 +33,25 @@ export type ArtistProfile = TimeField & {
...
+    address: controlledEthereumAddressFromId(artistProfile.artistId),
+    avatarUrl: artistProfile.avatarUrl,
     slug: slugify(`${artistProfile.name} ${artistProfile.createdAtTime.getTime()}`),
-    id: artistProfile.artistId,
     createdAtTime: artistProfile.createdAtTime,
     createdAtEthereumBlockNumber: artistProfile.createdAtEthereumBlockNumber
   }
 };
+
+export const distinctEarliestArtistProfiles = (artistProfiles: ArtistProfile[]): ArtistProfile[] => {
```
musnit, line 46: i think we said we'd go with a simple solution that just handles updating undefined fields, rather than caring about which is earliest?

puffo: There was unfortunately a few things that led me to just try to tackle it. While running the spindexer using the update-only solution, I realised that a batch of artist profiles sometimes contains two profiles for the same artist. 

Before running the db insert, we need to de-dup the set of artists (otherwise the "undefined-fields-only" insert statement breaks). I then found that de-duping using `uniq` or lodash's `uniqueBy` led to non deterministic choice of which artist profile is then used to seed the artist, which leads to different times, blocknumbers, and subsequently, different slugs.

So I thought that taking a similar approach to the approach we needed to take in the migration file would get us a step closer to a more reliable solution.

src/utils/identifiers.ts
puffo: Couldn't find a good function but I was looking in web3js. That from ethers should work.

puffo: I wasn't 100% confident that burn addresses are inaccessible wallets. I can remove though.

src/types/artist.ts
musnit: hm yeah i imagine it may make sense to just accept the non-determinism to make things easier, but on some fields (eg: slugs) it could be a significant issue

src/processors/default/createProcessedTracksFromAPI.ts
```diff
@@ -27,15 +27,15 @@ const processorFunction = (platformId: string, name: string) => async (apiTracks
...
+  const artists = distinctEarliestArtistProfiles(artistProfiles).map(profile => mapArtist(profile));
 
   // Because we assume an artist has the same ID across all profiles, we
   // don't need to worry about changing the artistId and artist{} fields in the processed
   // tracks and can just do a simple merge.
   const processedTracks = apiTracks.map(apiTrack => platformClient.mapAPITrack(apiTrack));
   const { mergedProcessedTracks } = await mergeProcessedTracks(processedTracks, clients.db, false);
 
-  await clients.db.upsert(Table.artists, artists);
+  await clients.db.insert(Table.artists, artists, { updateUndefinedOnConflict: 'id' });
```
musnit, line 38: isn't this still a upsert rather than an insert?

src/types/artist.ts
musnit: this logic still doesn't solve it fully though, for example:

 - aidan runs the indexer as is, indexes snoop dog from soundxyz, then aidan merges in a pull request that includes a new platform with some of snoop dogs much older nfts from a past platform. aidan gets the soundxyz slug
 - gian runs the indexer anew, after the above PR is merged. gian gets that oldplatform slug

even more so, I could imagine that even with the same PR, it's possible that 2 different people running a node get a different batch of nfts in their processPlatformTracks loop, one which includes snoop_sound and snoop_oldplatform an snoop_sound

so the nondeterminism emerges anyway

musnit: we could have a distinctEarliestArtistProfiles kind of logic applied at the database upsert level (like, with a CrdtUpsert) - may be worth considering if we can utilize that since it also would make use of functionality we already have

but even so, we're still left with the issue that a slug may change. maybe that's something we should just accept

src/db/migrations/44-add-artist-fields.ts
```diff
@@ -0,0 +1,60 @@
...
+        _.groupBy(withBlock, 'id'),
+        values => _.sortBy(values, 'createdAtBlockNumber')[0]
+      )
+    );
+
+    const updates = earliestDistinctUpdate.concat(withoutBlock).map((row: any) => {
+      return {
+        id: row.id,
+        address: ethereumAddressFromId(row.id),
+        avatarUrl: row.avatarUrl,
```
musnit, line 41: let's make sure we check later potential profiles for the avatarUrl in case the first one is missing, eg, see how the frontend does it here: https://github.com/Snowfork/spinamp/blob/e2d315afb41f1a0b0b3ffca00f080d7aafe0f0d8/src/utils/artists.ts#L9

src/types/artist.ts
puffo: From our chats, we're going to accept the problem with slugs changing or diverging. Last thing to do here is remove this extra logic. And also move the insert logic into upsert, along with the de-duplication.

src/db/sql-db.ts
```diff
@@ -82,6 +82,22 @@ const init = async (): Promise<DBClient> => {
           chunks.map(async chunk => {
             if (options?.ignoreConflict){
               await transaction.insert(chunk).into(tableName).onConflict(options.ignoreConflict as any).ignore();
+            } else if (options?.updateUndefinedOnConflict){
+              // create custom insert statement to only update `undefined` fields on conflict, for example:
+              // INSERT ... ON CONFLICT (id) DO UPDATE
+              //   SET field1 = COALESCE(field1, EXCLUDED.field),
+              //       field2 = COALESCE(field2, EXCLUDED.field2);
+              //
+              const setUndefinedFields = Object.keys(chunk[0] as any).map(key =>
```
puffo, line 91: Double check whether this is a homogenous chunk of keys

PR #130, opened 2022-11-14
update sound.xyz query

Geimaj:


src/clients/sound.ts
```diff
@@ -143,15 +143,15 @@ const init = async () => {
       return !!matchedNFT;
     });
     filteredAPITracks.forEach(apiTrack => {
-      if (apiTrack.tracks.length > 1) {
+      if (apiTrack.track.length > 1) {
```
musnit, line 146: @Geimaj did u confirm this is still an array right, even though the name is now singular?

Geimaj: This is actually returning an object not an array now, so I added the necessary changes

PR #131, opened 2022-11-15
blocks cache

Geimaj:


src/clients/blocks.ts
```diff
@@ -1,27 +1,48 @@
...
+      console.log(`using ${existingBlocks.length} blocks from cache`)
+      const existingBlocksByBlockNumber = _.keyBy(existingBlocks, 'blockNumber');
+
+      const missingBlockNumbers = blockNumbers.filter(blockNumber => !existingBlocksByBlockNumber[blockNumber]);
+      let newBlocks: Block[] = [];
 
       // only using the graph for Ethereum. All other EVM chains will use RPC calls
       if (chainId === ChainId.ethereum){
-        const blockNumbersForQuery = JSON.stringify(blockNumbers);
+        const blockNumbersForQuery = JSON.stringify(missingBlockNumbers);
```
musnit, line 45: should we deprecate the graph dependency entirely with this PR, even for Ethereum?

Geimaj: yeah I think we should! It would be nice to remove that dependency, and have a single code path for all chains. This change should reduce any performance impacts related to doing a query per block

src/clients/blocks.ts
```diff
@@ -1,49 +1,65 @@
...
+      let newBlocks: Block[] = [];
+
+      const blockNumbersParsed: number[] = missingBlockNumbers.map(blockNumber => parseInt(blockNumber))
+
+      const blocks = await evmClients[chainId].getBlockTimestampsByBlockNumber(blockNumbersParsed);
+
+      newBlocks = Object.keys(blocks).map(blockNumber => ({
+        chainId,
+        blockNumber,
+        timestamp: new Date(parseInt(blocks[blockNumber]) * 1000)
```
puffo, line 48: I didn't know about this. Does it mean that the block number officially corresponds to a distinct point in time, or are we making a bit of an approximation here?

puffo (approved):

src/clients/blocks.ts
```diff
@@ -1,49 +1,65 @@
-import { gql, GraphQLClient } from 'graphql-request';
 import _ from 'lodash';
 
+import { DBClient, Table } from '../db/db';
+import { Block } from '../types/block';
 import { ChainId } from '../types/chain';
 
 import { EVMClient } from './evm';
 
-const blocksAPI = new GraphQLClient(process.env.BLOCKS_SUBGRAPH_ENDPOINT!);
```
puffo, line 8: Can we remove the entry from .env.example too?

src/clients/blocks.ts
musnit: nah i think just badly named variables - blocks should be blockTimestamps?

src/clients/blocks.ts
```diff
@@ -1,49 +1,65 @@
...
+    ): Promise<{ 
+      timestampsByBlockNumber: { [blockNumber: string]: string },
+      newBlocks: Block[] 
+    }> => {
+      let timestampsByBlockNumber: { [blockNumber: string]: string } = {};
+
+      const existingBlocks = await dbClient.getRecords<Block>(Table.blocks, [
+        ['where', ['chainId', chainId]],
+        ['whereIn', ['blockNumber', blockNumbers]]
+      ])
```
musnit, line 34: hm so there will usually be 1000 or so blocks in this whereIn query? I guess that should be fine? as long as query pretty much instant?

musnit: we got 120k transfers, so first run probably 120 loops of this processor

src/clients/evm.ts
```diff
@@ -161,18 +161,19 @@ const init = async (providerUrl: string): Promise<EVMClient> => {
       const owner = await contract.owner();
       return owner
     },
-    getBlockTimestampsByBlockNumber: async (blockNumbers: number[]) => {
+    getBlockTimestampsByBlockNumber: async (blockNumbers: number[]): Promise<{ [blockNumber: string]: string }> => {
       const getBlock = provider.getBlock.bind(provider);
       const results = await rollPromises(blockNumbers, getBlock);
```
musnit, line 166: hmm so likely up to 1000 api calls to the endpoint? i think alchemy may throttle us but lets try and see. i guess with 120k+ transfers the first index could take quite long, but probably still fine if its first time only, especially if there are duplicates.

src/clients/blocks.ts
```diff
@@ -1,49 +1,65 @@
...
+
+      const blocks = await evmClients[chainId].getBlockTimestampsByBlockNumber(blockNumbersParsed);
+
+      newBlocks = Object.keys(blocks).map(blockNumber => ({
+        chainId,
+        blockNumber,
+        timestamp: new Date(parseInt(blocks[blockNumber]) * 1000)
+      }))
+
+      timestampsByBlockNumber = blocks
```
musnit, line 51: can just create this variable here rather than at the start?

src/clients/blocks.ts
Geimaj: Yeah usually 100 blocks because we slice it to that length. This doesn't change how many times the processor will run, just how many requests will be sent. The queries are pretty much instant, and after this has run a few times the number of requests drops significantly as we don't fetch duplicate blocks and just read those from the DB

src/clients/evm.ts
Geimaj: It's been okay with my testing so far, but if we do run into trouble this will be easy to limit since we are already using rollPromises

musnit (approved):

PR #132, opened 2022-11-16
Geimaj/identities cache

Geimaj:
TODO:

- [ ] index ens names
- [ ] index ens avatar
- [ ] update artist names
- [ ] remove lens profile indexing from lens metaFactory and replace with this

Notes:
I had to make changes to the getEVMEvents function. This is mainly because it was designed to query a set of contracts, but in this case I need to query a single contract with a set of parameters. This complicates how the cursor is managed and I still need to test it against the rest of the codebase.

Another change was adding a limit, so we always fetch {limit} oldest items from the cursor as 'mostStaleContracts' - this is to split the fetch into different requests, to avoid having a topic with like 40k+ or arguments. 

I still need to rename 'staleContracts' in the newEVMEvents to 'staleArgs', as now this works with any set of arguments, and not just a set of contract addresses.


src/triggers/newEVMEvent.ts
```diff
@@ -0,0 +1,141 @@
...
+  updateCursor?: (rangeEnd: string, staleContractFilters: ContractFilter[]) => { [x: string]: string },
+  limit?: number
+) => Trigger<Cursor> = (
+  chainId: ChainId,
+  args: EVMEventCursorArgs[],
+  contractFilters: ContractFilter[],
+  gap: string = process.env.ETHEREUM_BLOCK_QUERY_GAP!,
+  staleFilter,
+  updateCursor,
+  limit = 5000
```
Geimaj, line 59: add a limit to get the last {limit} arguments from the cursor, to prevent RPC queries with a huge number of OR conditions

src/triggers/newEVMEvent.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    const newEvents = await client.getEventsFrom(
+      rangeStart.toString(),
+      rangeEnd.toString(),
+      staleContractFilters,
+    );
+    const newCursor = JSON.stringify(newCursorObject);
+    // if (newEvents.length === 0 && mostStaleContracts.length !== args.length) {
+    if (newEvents.length === 0 && mostStaleContracts.length !== limit) {
```
Geimaj, line 129: Need to test this recursion condition further

src/db/migrations/45-add-identities.ts
```diff
@@ -0,0 +1,24 @@
+import { Knex } from 'knex';
+
+import { Table } from '../db';
+import { updateViews } from '../migration-helpers';
+
+
+export const up = async (knex: Knex) => {
+
+  await knex.schema.createTable(Table.identities, table => {
```
musnit, line 9: q - should we call this Table.addresses or identities?
 - since address is the primary key, I'd think Table.addresses should be the main
 - we already support multiple addresses on an identity in our other user profile API, and I imagine we'll wanna keep doing this. especially if we wanna converge these 2 databases at some point, I think Table.addresses is closer to that?
 - having said that, name/avatar are more like profile/identity fields rather than address fields

I think for now keeping it as Table.addresses likely makes sense, and just keeping the other fields on this same table for now for simplicity is fine too?

Also maybe it should replace the collectors table too, or we just rename the collectors table to Table.addresses?

src/processors/default/populateLensAvatar.ts
```diff
@@ -0,0 +1,76 @@
...
+  return output
+} 
+
+
+export const populateLensAvatar: Processor =
+{
+  name: NAME,
+  trigger,
+  processorFunction: async (triggerOutput , clients: Clients) => {
+    const updatedItentites: Identity[] = triggerOutput.items.map((event: ethers.Event) => {
```
puffo, line 66:
```suggestion
    const updatedIdentites: Identity[] = triggerOutput.items.map((event: ethers.Event) => {
```

src/triggers/newEVMEvent.ts
puffo: Ping me if you'd like to pair on setting up some codebase tests :)

PR #133, opened 2022-11-17
Adjust seeds API to support artists changing details

puffo:


src/seeds/access.ts
```diff
@@ -38,17 +39,30 @@ export const restrictAccess = (payload: MessagePayload, signer?: EthereumAddress
   }
 }
 
-export const ownerOrAdmin = (payload: MessagePayload, signer?: EthereumAddress) => {
+export const ownerOrAdmin = async (payload: MessagePayload, signer: EthereumAddress, dbClient: DBClient) => {
   if (!signer) {
     throw new Error('must specify a signer');
   }
 
   if (!notAdminAddress(signer)) {
```
musnit, line 47: double not is usually bad for readability lol - maybe change the function to isAdminAddress ?

src/seeds/access.ts
```diff
@@ -38,17 +39,30 @@ export const restrictAccess = (payload: MessagePayload, signer?: EthereumAddress
...
-    throw new Error(`only owner or admin can ${payload.operation} ${payload.entity}`);
+  const persistedRecords: any = await dbClient.getRecords(payload.entity as Table, [
+    ['where', [ 'id', payload.data.id ], ],
+    ['where', [ 'address', payload.data.address ], ],
+  ])
+
+  if (!persistedRecords[0] || payload.data?.address !== signer) {
+    // throw new Error(`only owner or admin can ${payload.operation} ${payload.entity}`);
+    throw new Error('not found');
   }
```
musnit, line 66: should move into a separate function with an informative name?

src/seeds/access.ts
```diff
@@ -2,22 +2,66 @@ import { isValidChecksumAddress } from 'ethereumjs-util';
...
+  if (!signer) {
+    throw new Error('must specify a signer');
+  }
+
+  if (!notAdminAddress(signer)) {
     return;
   }
-  onlyAdmin(signer);
+
+  if (payload.data?.address !== signer) {
```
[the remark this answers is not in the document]
puffo: [fd3adc9](https://github.com/spinamp/spindexer/pull/133/commits/fd3adc9981d47897bdc10fb41041ec728a442509) adjusted this to do an `id` and `address` lookup, effectively ensuring that the signer has scope on the artist ID only when the signer-address matches the payload address. 

I will definitely triple check this though!

PR #134, opened 2022-11-21
Geimaj/lens follow up

Geimaj:
adds more lens songs and gets audio from `media` rather than `animation_url`


src/types/media.ts
```diff
@@ -7,6 +7,13 @@ export enum MimeEnum {
   jpeg = 'image/jpeg',
   png = 'image/png',
   pdf = 'application/pdf',
+  quicktime = 'video/quicktime'
+}
+
+export type MediaMetadata = {
```
musnit, line 13:
```suggestion
export type LensMediaMetadata = {
```

src/types/nft.ts
```diff
@@ -145,3 +146,25 @@ export const getTraitType = (nft: NFT, value: string) => {
   }
   return traitAttribute.trait_type;
 };
+
+export const getMedia = (nft: NFT, mimeTypes: MimeEnum[]) => {
```
musnit, line 150:
```suggestion
export const getLensMedia = (nft: NFT, mimeTypes: MimeEnum[]) => {
```

src/types/fieldExtractor.ts
```diff
@@ -105,6 +107,7 @@ const titleExtractors: TitleExtractorMapping = {
 const audioUrlExtractors: AudioUrlExtractorMapping = {
   [AudioUrlExtractorTypes.METADATA_ANIMATION_URL]: (nft: NFT) => cleanURL(nft.metadata.animation_url),
   [AudioUrlExtractorTypes.ATTRIBUTES_TRAIT_AUDIO]: (nft: NFT) => cleanURL( getTrait(nft, 'Audio')),
+  [AudioUrlExtractorTypes.FIND_AUDIO_MEDIA]: (nft: NFT) => cleanURL(getMedia(nft, [MimeEnum.mp3, MimeEnum.wav, MimeEnum.xWav, MimeEnum.mp4, MimeEnum.quicktime]))
```
musnit, line 110: can move the array into types/media as something like AudioAndVideoTypes ?

src/db/migrations/44-lens-approvals.ts
```diff
@@ -0,0 +1,69 @@
+import { Knex } from 'knex';
+
+import { formatAddress } from '../../types/address';
+import { ChainId, getFactoryId } from '../../types/chain';
+import { CrdtOperation } from '../../types/message';
+import { Table } from '../db';
+
+const lensCurated = [
+  '0x727Ce17C09a74292Eb1594f70e8f048402b6DC02',
```
musnit, line 9: i see 39 tracks here but a bunch less on the preview @ https://spindexer-qa-build.spinamp.xyz/platform/lens

are some broken?

src/db/migrations/44-lens-approvals.ts
```diff
@@ -0,0 +1,69 @@
...
+      approved: true,
+      autoApprove: true,
+    },
+    operation: CrdtOperation.UPDATE,
+  }));
+
+  for (const message of messages) {
+    await knex(Table.seeds).insert(message);
+  }
+};
```
puffo, line 56: I think we should start to see if it's worthwhile trying out the new seed process which won't leave us with migrations to clean up later. We can also reflect on whether it is worth the extra steps or not :)

https://www.notion.so/Seed-process-for-Artist-Integrations-46e283d316644dfc92f413f44f8eae77

puffo (approved):

src/db/migrations/44-lens-approvals.ts
musnit: @puffo yeah there's a task here assigned to u to migrate over to it when you're done with your current work: https://linear.app/spinamp/issue/SPI-676/deprecate-seed-migrations-properly

src/db/migrations/44-lens-approvals.ts
Geimaj: Just had to remove a few duplicate values

PR #138, opened 2022-11-23
Add trigger+processor for fetching content type from IPFS

puffo:


src/db/migrations/41-add-signer-to-mempool.ts
```diff
@@ -17,5 +18,5 @@ export const down = async (knex: Knex) => {
     table.dropColumn('signer');
   });
 
-  await updateViews(knex);
+  await updateViews(knex, overridesV1);
```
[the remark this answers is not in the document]
musnit: maybe we can just delete all these updateViews and only keep it on the last one?

src/types/track.ts
```diff
@@ -13,6 +13,11 @@ export type NFTTrackJoin = {
   processedTrackId: string;
 }
 
+export enum SourceIPFS {
+  AUDIO = 'Audio',
+  ARTWORK = 'Artwork'
+}
```
musnit, line 19: maybe just keep this within the processor/trigger since it's kinda specific just to those?

src/triggers/missing.ts
```diff
@@ -42,6 +43,21 @@ export const missingMetadataIPFSHash: Trigger<undefined> = async (clients) => {
   return nfts;
 };
 
+export const missingMimeType: (source: SourceIPFS) => Trigger<undefined> = (source) => {
```
musnit, line 46: lets put this in a folder with the processor. actually in general i think we should be keeping triggers together with their processors in the same file or folder, and refactor towards that direction since there's very little sharing of triggers across processors.

src/processors/default/addMimeTypeToProcessedTracks.ts
```diff
@@ -0,0 +1,33 @@
...
+        const id = processedTrack.id;
+        const ipfsHash = processedTrack[`lossy${source}IPFSHash`];
+        const response = await clients.axios.head(`${process.env.IPFS_ENDPOINT}${ipfsHash}`, { timeout: parseInt(process.env.METADATA_REQUEST_TIMEOUT!) })
+
+        const result: any = { id: id }
+        result[`lossy${source}MimeType`] = response.headers['content-type'];
+        return result;
+      }
+
+      const results = await rollPromises<ProcessedTrack, any, void>(
```
musnit, line 24: need error handling, and probably could use same error/retry functionality that we have for uploading ipfs files? else this could retry indefinitely for a broken file/link?

src/processors/default/addMimeTypeToProcessedTracks.ts
```diff
@@ -0,0 +1,33 @@
...
+    processorFunction: async (processedTracks: ProcessedTrack[], clients: Clients) => {
+      console.log(`Fetching processed track ${source} mime types`);
+
+      const updatedMimeTypes = async (processedTrack: any) => {
+        const id = processedTrack.id;
+        const ipfsHash = processedTrack[`lossy${source}IPFSHash`];
+        const response = await clients.axios.head(`${process.env.IPFS_ENDPOINT}${ipfsHash}`, { timeout: parseInt(process.env.METADATA_REQUEST_TIMEOUT!) })
+
+        const result: any = { id: id }
+        result[`lossy${source}MimeType`] = response.headers['content-type'];
```
musnit, line 20: should we do some basic sanity check/validation on the response content-type? probably maintain a list of acceptable mime type results and error if it doesn't match one of those? and toLowerCase()

puffo: Also extend the mime types to those supported by cloudinary

src/processors/default/addMimeTypeToProcessedTracks.ts
```diff
@@ -0,0 +1,92 @@
...
+        try {
+          response = await clients.axios.head(`${process.env.IPFS_ENDPOINT}${ipfsHash}`, { timeout: TIMEOUT })
+          contentType = response.headers['content-type']?.toLowerCase();
+        } catch (e: any) {
+          errorMsg = `Error: failed to fetch ${source} mime type for ipfs hash: ${ipfsHash} with error: ${e.message}`;
+        }
+
+        if (contentType && !Object.values(MimeEnum).includes(contentType)) {
+          errorMsg = `Error: invalid ${source} mime type: ${contentType} for ipfs hash: ${ipfsHash}`;
+        }
```
musnit, line 66: i think we should explicitly check that:
 - AUDIO is audiovisual (ie, not an image)
 - ARTWORK is image/visual (ie, not pure audio)

src/processors/default/addMimeTypeToProcessedTracks.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type TrackNftJoin = ProcessedTrack & Partial<NFT>;
+
+const TIMEOUT = parseInt(process.env.METADATA_REQUEST_TIMEOUT!)
+const QUERY_LIMIT = process.env.QUERY_TRIGGER_BATCH_SIZE
+
+const missingMimeType: (source: SourceIPFS) => Trigger<undefined> = (source) => {
+  return async (clients) => {
+    // find approved tracks (and the nftID) which are missing a mimeType; skipping any which have had errors
+    const query = `select p.*, n.id as "nftId"
```
musnit, line 24: hmm not sure it's a good idea to use the nft ID here - especially since:
 - there are usually 20+ nfts per track so with this outer join, there could be 25+ TrackNftJoin items for each track and so we'd repeat this 25 times
 - feels bad to use the nft process errors table for this, since again we'd be storing duplicate errors for every track and need to worry about the association between tracks and nfts just adding unnecessary complexity to this code

I do imagine in future we'd potentially want to check mimetype on nfts directly, rather that on tracks, but even then i don't think involving a relation here is good design.

I see the intention here is to use existing nft errors table, but feels over complex/messy and suggest we implement it with a cleaner trigger, so some options:
 a - Create a new processTrackErrors table for this
 b - Improve the nftErrors table to be more generic and support errors across different tables/data types
 c - Use the ipfsFiles table (which also has error fields in it) for this.

(c) looks easiest/simplest for now, and most appropriate since most the logic in this PR is actually ipfsFile specific, nevermind even track specific. I think implementing this to be ipfsFile specific rather than just track will also serve us well in future _(eg: if we want to process ipfsFiles on nfts or potentially videos and other things in future)_

So an idea to with this would be:
 - Actually store mimeType per file on the ipfsFiles
 - Similar trigger to those in `src/triggers/ipfs.ts` that triggers on records with null error and null mimeType
 - Processor stores both mimeType and errors on the ipfs file record, similar to how existing ipfsMediaUploader processor works
 - Above is great, now we have mimetypes on ipfsFiles, but still need to link them to tracks, so for that:
    - maybe just update the processedTracks view to join on ipfsFiles and include the mimetypes as part of the view and filter out null mimetypes in the view

src/processors/default/addMimeTypeToProcessedTracks.ts
musnit: in the context of my other comment (https://github.com/spinamp/spindexer/pull/138/files#r1032509476), if we go with that solution, i think we should either:
 - have extra fields on the ipfsFiles table like isAudio and isVideo and isImage that we store and use to match against as part of the tracks view
 - match against the hardcoded array of mimetypes as part of the view to filter out mismatched audio/artwork

puffo: @musnit just to make sure I understand correctly, the "matching" we're talking about is just updating the views for `processed_tracks` right?

At the moment, `processed_tracks` includes both `lossyArtworkIPFSHash` and `lossyAudioIPFSHash` where both are not null, so we'd update the `processed_tracks` view to ensure that we:

1. exclude `lossyArtworkIPFSHash` and `lossyAudioIPFSHash` that do not have a mime type
2. only return `processed_tracks` where `lossyArtworkIPFSHash` is image/video AND `lossyAudioIPFSHash` is audio/video

Is that in-line with what you're thinking?

musnit: 1 - yes, though exclude the whole track if either is missing a mime type, not just those fields
2 - yes

and yeah this sounds good. I suppose there may be some edge cases we uncover where we still actually want to return the track, but let's implement it like this and then see

src/processors/default/ipfsMimeTypeProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+import { IPFSFile } from '../../types/ipfsFile';
+import { MimeEnum } from '../../types/media';
+import { Clients, Processor } from '../../types/processor';
+import { Trigger } from '../../types/trigger';
+import { rollPromises } from '../../utils/rollingPromises';
+
+const TIMEOUT = parseInt(process.env.METADATA_REQUEST_TIMEOUT!)
+const QUERY_LIMIT = process.env.QUERY_TRIGGER_BATCH_SIZE
+
+const missingMimeType: Trigger<undefined> = async (clients) => {
```
musnit, line 12: looks like the current ipfs uploader is actually being limited by the IPFS_UPLOAD_BATCH_SIZE env variable which i have set to just 1

so i'd suggest maybe add another env variable for your HEAD getter that's lower than the global limits and go down to like 10 max items on the trigger

also, trigger only on files after they've been pinned

src/processors/default/ipfsMimeTypeProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+        let contentType: any = '';
+
+        try {
+          response = await clients.axios.head(`${process.env.IPFS_ENDPOINT}${ipfsHash}`, { timeout: TIMEOUT })
+          contentType = response.headers['content-type']?.toLowerCase();
+        } catch (e: any) {
+          errorMsg = `Error: failed to fetch mime type for ipfs hash: ${ipfsHash} with error: ${e.message}`;
+        }
+
+        // todo: enrich with isAudio, isVideo, isImage flags
```
musnit, line 43: todo

src/db/views.ts
```diff
@@ -1,18 +1,26 @@
...
+
+    join "${Table.ipfsFiles}" f
+    on t."lossyAudioIPFSHash" = f.cid
+    join "${Table.ipfsFiles}" f1
+    on t."lossyArtworkIPFSHash" = f1.cid
+
     where "lossyArtworkIPFSHash" is not null
     and "lossyAudioIPFSHash" is not null
+    and f."mimeType" is not null
+    and f1."mimeType" is not null
```
musnit, line 23: add extra filters where f."mimeType" is one of audio/video and f1 is image/video

src/db/views.ts
```diff
@@ -1,18 +1,26 @@
...
+
     join "${Table.ipfsPins}" p
     on t."lossyAudioIPFSHash" = p.id
     join "${Table.ipfsPins}" p1
     on t."lossyArtworkIPFSHash" = p1.id
+
+    join "${Table.ipfsFiles}" f
+    on t."lossyAudioIPFSHash" = f.cid
+    join "${Table.ipfsFiles}" f1
+    on t."lossyArtworkIPFSHash" = f1.cid
```
musnit, line 18:
```suggestion
    join "${Table.ipfsFiles}" audio_file
    on t."lossyAudioIPFSHash" = audio_file.cid
    join "${Table.ipfsFiles}" artwork_file
    on t."lossyArtworkIPFSHash" = artwork_file.cid
```

src/processors/ipfsFile/mimeTypeProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+        const result = ipfsFile;
+
+        if (errorMsg) {
+          result.error = errorMsg;
+        } else {
+          result.mimeType = contentType;
+        }
+
+        return result;
+      }
```
musnit, line 57: can we declare this in the media type file or some utility file and export the function so that we can use it in other places too?

src/db/migrations/27-recreate-views.ts
```diff
@@ -1,10 +1,9 @@
 
 import { Knex } from 'knex'
 
-import { updateViews } from '../migration-helpers'
 
 export const up = async (knex: Knex) => {
-  await updateViews(knex);
+  // old code recreated views
```
musnit, line 6:
```suggestion
  // this is deprecated _(old code recreated views)_
```

src/types/media.ts
```diff
@@ -33,3 +57,35 @@ export const isGif = (mimeType?: MimeEnum) => {
...
+    contentType = response.headers['content-type']?.toLowerCase();
+  } catch (e: any) {
+    errorMsg = `Error: failed to fetch mime type for ipfs hash: ${ipfsHash} with error: ${e.message}`;
+  }
+
+  if (contentType && !Object.values(MimeEnum).includes(contentType)) {
+    errorMsg = `Error: unsupported mime type '${contentType}' for ipfs hash: ${ipfsHash}`;
+  }
+
+  const updated = ipfsFile;
```
musnit, line 79: Let's not change the object that's passed in as a parameter, since then this loses it's purity:

```suggestion
  const updated = { ...ipfsFile };
```

PR #140, opened 2022-11-29
Allow many URLs for the same CID

puffo:


src/processors/ipfsFile/mediaUploader.ts
```diff
@@ -0,0 +1,96 @@
...
+import { IPFSFile, IPFSFileUrl } from '../../types/ipfsFile';
+import { Clients, Processor } from '../../types/processor';
+import { ProcessedTrack } from '../../types/track';
+import { Trigger } from '../../types/trigger';
+import { rollPromises } from '../../utils/rollingPromises';
+
+type TrackUrlJoin = ProcessedTrack & Partial<IPFSFileUrl>;
+type MediaType = 'Artwork' | 'Audio';
+
+const notOnIpfs: (media: MediaType) => Trigger<undefined> =
```
musnit, line 12: Could we make this more generic (especially to support other tables, so that we can use it for the other outstanding artist profile picture PR)

ie, change this from MediaType to

```
RecordUrl = {
  table: Table,
  urlField: string,
  cidField: string
}
```

src/processors/ipfsFile/mediaUploader.ts
```diff
@@ -0,0 +1,96 @@
...
+(media) => async (clients: Clients) => {
+  const query = `select * from "${Table.processedTracks}" as t
```
musnit: can we select just the fields we need, instead of * ?

musnit: I guess we may need something like `idFields: string[],` in the above RecordUrl too so that we know which field to select for the record id(s)

src/processors/ipfsFile/mediaUploader.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      delete trackWithUrl['url']
+      delete trackWithUrl['cid']
+      delete trackWithUrl['error']
+      delete trackWithUrl['lastRetry']
+      delete trackWithUrl['numberOfRetries']
+
+      const track: ProcessedTrack = {
+        ...trackWithUrl
+      }
```
musnit: Instead of creating this track update with all the fields and then deleting fields we don't need, can we just create the record update using only the exact fields we need (ie, id(s) and cidField only)?

The deletion flow is not future-compatible, since it basically assumes that once the extra fields are deleted, the update remaining is minimal, but if we change/add fields in future, this could break. An exact-field creation flow should not introduce this implicit assumption/risk.

src/processors/ipfsFile/mediaUploader.ts
```diff
@@ -0,0 +1,96 @@
...
+    await clients.db.update(Table.processedTracks, trackUpdates)
+    await clients.db.upsert(Table.ipfsFiles, ipfsFiles, 'cid');
+    await clients.db.upsert(Table.ipfsFilesUrls, ipfsFilesUrls, 'url');
+  }
+}
+
+export const ipfsMediaUploader: (media: MediaType) => Processor =
+(media) => {
+  return {
+    name: `ipfs${media}Uploader`,
```
musnit, line 78:
```suggestion
    name: `ipfsUploader:${media.table}:${media.urlField}`,
```

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -11,29 +12,80 @@ const ipfsFilesOutOfSyncWithPins: (field: TrackMediaField) => Trigger<undefined>
...
     left outer join "${Table.ipfsFiles}" as file
     on t."${field}IPFSHash" = file.cid
-    where
+    where (
       t."${field}IPFSHash" is not null and
       file.cid is null and
       file.error is null and
       t."${field}URL" ILIKE ('%' || t."${field}IPFSHash" || '%')
+    ) or (
+      t."${field}IPFSHash" is not null and
```
musnit: not sure i understand what you're trying to do here - this or condition looks like it will return a superset of all the records, essentially making the previous clause irrelevant?

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -11,29 +12,80 @@ const ipfsFilesOutOfSyncWithPins: (field: TrackMediaField) => Trigger<undefined>
...
+    name: 'ipfsFileSyncExternalPinsProcessor',
+    trigger: ipfsFilesPinnedWithoutUrl(field),
+    processorFunction: async (input: any[], clients: Clients) => {
+      console.log(`Adding ${field} ipfs files for media pinned without URLS`);
+
+      const tracks: Partial<ProcessedTrack>[] = input.map((row: any) => ({
+        id: row.id,
+        [`${field}URL`]: `${process.env.IPFS_ENDPOINT}${row[`${field}IPFSHash`]}`,
+      }))
+      await clients.db.update(Table.processedTracks, tracks);
```
musnit: can we move all the updates together at the end plz

src/db/migrations/49-extract-ipfs-files-urls-from-ipfs-files.ts
```diff
@@ -0,0 +1,60 @@
...
+      urlCids.push({ cid: r.cid, url: r.url })
+    });
+
+    // remove url from ipfsFiles table
+    const { url, ...rest } = row[0];
+    return rest;
+  });
+
+  // repopulate ipfsFiles table
+  await knex(Table.ipfsFiles).truncate();
```
musnit, line 33: so this deletes all records in the table, and then re-inserts them without the url?

this seems fine, though above u only get records where "cid" is not null. so the other records will be lost? i guess that should be fine?

src/processors/ipfsFile/errorProcessor.ts
```diff
@@ -29,7 +43,22 @@ export const ipfsFileErrorRetry: Processor = {
...
+  },
+  initialCursor: undefined
+}
+
+export const ipfsFileUrlErrorRetry: Processor = {
+  name: `ipfsFileUrlErrorRetry`,
+  trigger: ipfsFilesUrlsWithErrors,
+  processorFunction: async (input: IPFSFileUrl[], clients: Clients) => {
+    const errorUpdates: IPFSFileUrl[] = input.map((ipfsFileUrl) => ({
+      ...ipfsFileUrl,
```
musnit, line 56: do we need all the fields with ...? can't we just set the primary key, ie, url and do an update?

src/processors/ipfsFile/errorProcessor.ts
```diff
@@ -29,7 +43,22 @@ export const ipfsFileErrorRetry: Processor = {
...
+      error: undefined,
+      numberOfRetries: (ipfsFileUrl.numberOfRetries ?? 0) + 1,
+      lastRetry: new Date()
+    }));
+    await clients.db.upsert(Table.ipfsFilesUrls, errorUpdates, 'url', undefined, true);
```
musnit, line 61: we really need to stop using upsert when we mean update - feels like this issue is just growing and we're spreading tech debt - maybe should fix the update function to actually work properly?

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -11,29 +12,80 @@ const ipfsFilesOutOfSyncWithPins: (field: TrackMediaField) => Trigger<undefined>
...
         url: row[`${field}URL`],
         cid: row[`${field}IPFSHash`],
       }))
-      await clients.db.insert(Table.ipfsFiles, files);
+      await clients.db.insert(Table.ipfsFilesUrls, filesUrls);
+    }
+  }
+}
+
+export const ipfsFileSyncExternalPinsProcessor: (field: TrackMediaField) => Processor =
```
musnit: this processor is related to ExternalUploads, not ExternalPins - let's rename it and the trigger to refer to uploads rather than pins to avoid confusion

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -11,29 +12,80 @@ const ipfsFilesOutOfSyncWithPins: (field: TrackMediaField) => Trigger<undefined>
...
   return {
     name: 'ipfsFileSyncExistingPinsProcessor',
     trigger: ipfsFilesOutOfSyncWithPins(field),
-    processorFunction: async (input: IPFSFile[], clients: Clients) => {
+    processorFunction: async (input: any[], clients: Clients) => {
       console.log(`Adding ${field} ipfs files for media pinned by other platforms`);
 
-      const files = input.map((row: any) => ({
+      const files: IPFSFile[] = input.map((row: any) => ({ cid: row[`${field}IPFSHash`] }))
+      await clients.db.insert(Table.ipfsFiles, files);
```
musnit: all db operations at end plz

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -11,29 +12,80 @@ const ipfsFilesOutOfSyncWithPins: (field: TrackMediaField) => Trigger<undefined>
...
+      t."${field}IPFSHash" is not null and
+      t."${field}URL" is null and
+      url.cid is null and
+      url.error is null
+  `)
+
+   return tracksWithCIDWithoutURL.rows;
+ }
+
 export const ipfsFileSyncExistingPinsProcessor: (field: TrackMediaField) => Processor =
```
musnit: change to ipfsFileSyncExistingUploadsProcessor rather than pins (and same for the trigger name/console.log to avoid confusion)

src/processors/ipfsFile/mediaUploader.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (!cid){
+          const file = await clients.ipfs.client.add(source, {
+            pin: false,
+            timeout: parseInt(process.env.IPFS_API_TIMEOUT!)
+          });
+          cid = file.cid.toString();
+        }
+
+        trackUpdates.push({
+          ...track,
```
musnit: don't use ... -> this is putting all fields into the update, even ones that may not be relevant to this processor. we should only put in specific fields we actually want to update

src/index.ts
```diff
@@ -87,13 +87,16 @@ const PROCESSORS = (
     addTimestampFromMetadata,
     ...platformTrackProcessors,
     ...apiTrackProcessors,
-    ipfsAudioUploader,
-    ipfsArtworkUploader,
-    ipfsAudioPinner,
+    ipfsFileUrlErrorRetry,
+    ipfsMediaUploader('Artwork'),
+    ipfsMediaUploader('Audio'),
```
musnit, line 96: does uploader stuff need to run after pin sync stuff? since otherwise we'll just be re-uploading unnecessarily anyway?

src/processors/ipfsFile/syncExistingPinsProcessor.ts
puffo: Yep, removing.

src/db/migrations/49-extract-ipfs-files-urls-from-ipfs-files.ts
puffo: Yep, those are the errored ones which then get repopulated on a reindex. I've verified with a local reindex 👍

src/processors/ipfsFile/errorProcessor.ts
puffo: Agreed. Problem is that our update doesn't support setting a field to undefined. Will add this to the changes on the other PR

src/types/track.ts
```diff
@@ -13,8 +13,8 @@ export type NFTTrackJoin = {
   processedTrackId: string;
 }
 
-export type ProcessedTrackAudio = { lossyAudioURL: string } | { lossyAudioIPFSHash: string };
-export type ProcessedTrackArtwork = { lossyArtworkURL: string } | { lossyArtworkIPFSHash: string };
+export type ProcessedTrackAudio = { lossyAudioURL: string } | { lossyAudioIPFSHash: string } | { lossyAudioMimeType: string };
+export type ProcessedTrackArtwork = { lossyArtworkURL: string } | { lossyArtworkIPFSHash: string } | { lossyArtworkMimeType: string };
```
musnit: whats this for?

src/processors/ipfsFile/errorProcessor.ts
musnit: cool plz add to linear task

src/processors/ipfsFile/syncExistingPinsProcessor.ts
```diff
@@ -1,39 +1,85 @@
...
 (field) => {
   return {
-    name: 'ipfsFileSyncExistingPinsProcessor',
-    trigger: ipfsFilesOutOfSyncWithPins(field),
-    processorFunction: async (input: IPFSFile[], clients: Clients) => {
-      console.log(`Adding ${field} ipfs files for media pinned by other platforms`);
+    name: 'ipfsFileSyncExistingUploadsProcessor',
+    trigger: ipfsFilesOutOfSyncWithUploads(field),
+    processorFunction: async (input: any[], clients: Clients) => {
+      console.log(`Adding ${field} ipfs files for media uploaded by other platforms`);
```
musnit:
```suggestion
      console.log(`Adding ${field} ipfs files for media already on IPFS`);
```

musnit:
closing - will be merged as part of parallelization refactor PR coming soon

PR #144, opened 2022-11-30
IPFS table becomes CID source of truth

puffo:
- [x] write tests to verify
- [x] generalise to handle artwork AND audio IPFS hashes
- [x] avoid triggering when ipfsFile is already populated


src/db/migrations/48-backfill-ipfs-files.ts
```diff
@@ -0,0 +1,32 @@
...
+    left outer join "${Table.ipfsFiles}" as i
+    on t."lossyArtworkURL" = i.url
+    where (
+      (t."lossyArtworkIPFSHash" is not null and t."lossyArtworkURL" is not null)
+      and
+      (t."lossyArtworkURL" ILIKE ('%' || t."lossyArtworkIPFSHash" || '%'))
+      and
+      (i.url is null and i.error is null)
+    )
+  `)
```
musnit, line 18: could you simplify this to something like:
```
select distinct "lossyArtworkIPFSHash" from "${Table.processedTracks}" as t
left outer join "${Table.ipfsFiles}" as i
on t."lossyArtworkIPFSHash" = i.cid
where i.cid is null
and i.error is null
```

musnit: ie, don't even worry about the urls in the join/check

musnit: i guess checking for matching url is important to avoid mismatch, yeh, but i think the insert should still be cid-centric, so maybe:
```
select distinct "lossyArtworkIPFSHash" from "${Table.processedTracks}" as t
left outer join "${Table.ipfsFiles}" as i
on t."lossyArtworkIPFSHash" = i.cid
where
t."lossyArtworkIPFSHash" is not null and
i.cid is null and
i.error is null and
t."lossyArtworkURL" ILIKE ('%' || t."lossyArtworkIPFSHash" || '%')
```

puffo:
Pair approved with @Geimaj

PR #147, opened 2022-12-09
feat: upload avatars to IPFS

xtyrrell:
Adds a processor to upload artist profiles' avatars to IPFS.

Caveat: uploaded files will not be pinned. We need to update the IPFS pinner processor because it currently only looks at the `processedTracks` table  (in `src/processors/default/ipfsPinner.ts`). It should look at the IPFS files table.

**Next steps**
- remove avatarIPFSHash column and instead only use entries in IPFS files table (will need to change trigger too)
- probably also want to update the view to join on this table
- use error processor for retries

### DIFF
--- a/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
+++ b/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
@@ -1,0 +1,16 @@
+export class IPFSUploadError extends Error {
+  readonly url: string;
+
+  constructor(message: string, url: string) {
+    super(message);
+
+    // 👇️ because we are extending a built-in class
+    Object.setPrototypeOf(this, IPFSUploadError.prototype);
+
+    this.url = url;
+  }
+
+  getErrorMessage() {
+    return 'Error when uploading file to IPFS: ' + this.message;
+  }
+}
--- a/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
+++ b/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
@@ -1,13 +1,17 @@
 export class IPFSUploadError extends Error {
+  readonly name = 'IPFSUploadError';
   readonly url: string;
 
   constructor(message: string, url: string) {
-    super(message);
+    super(getErrorMessage(message, url));
 
     // 👇️ because we are extending a built-in class
     Object.setPrototypeOf(this, IPFSUploadError.prototype);
 
     this.url = url;
   }
+}
 
+function getErrorMessage(url: string, message: string) {
+  return `Error when uploading file to IPFS at url "${url}": ${message}`;
 }
--- a/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
+++ b/src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
@@ -10,7 +10,4 @@ export class IPFSUploadError extends Error {
     this.url = url;
   }
 
-  getErrorMessage() {
-    return 'Error when uploading file to IPFS: ' + this.message;
-  }
 }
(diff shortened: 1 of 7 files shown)


src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    // 👇️ because we are extending a built-in class
+    Object.setPrototypeOf(this, IPFSUploadError.prototype);
+
+    this.url = url;
+  }
+
+  getErrorMessage() {
+    return 'Error when uploading file to IPFS: ' + this.message;
+  }
```
musnit: this function is never used, is it? when the error is caught won't it just use the super(message) without the url / prefix text? should this rather use something like `super(getErrorMessage(message, url))` ?

src/processors/uploadAvatarToIpfs/processor.ts
```diff
@@ -0,0 +1,130 @@
...
+    .map(error => ({url: error.url, error: error.message}));
+
+  const ipfsFileUpdates = [
+    ...ipfsFileSuccessUpdates,
+    ...ipfsFileFailureUpdates,
+  ];
+
+  if (ipfsFileFailureUpdates.length > 0) console.log("IPFS file uploads failed: ", JSON.stringify(ipfsFileFailureUpdates))
+
+  // TODO: update rather than upsert (needs change to db client to accept multiple IDs)
```
musnit: yeah. i think can just remove this comment for now since we're tracking this upsert issue elsewhere already

musnit:
code looks great, thanks! small comment, otherwise also re **Next Steps**:
 - uploaded files will not be pinned:
   - [ ] @puffo is working another PR that's reworking the ipfsFiles table to be the source of truth for all content. once that's completed, updating the IPFS pinner processor to target that table as you suggest sounds good. can maybe wait for that work to be completed then before merging this PR
 - remove avatarIPFSHash column:
   - the above PR is going to be changing the primary key on ipfsFiles so that the cid/hash is actually the primary key, so we'll still need this column as the foreign key
 - use error processor for retries
   - [ ] yeah may make sense to update this in this PR too
 - update the view to join on this table
   - we're gonna be re-working the way artistProfiles get merged together into a single artist record, so i think can hold off on this for now as that future PR should handle merging the avatar url up to the artist record properly

src/processors/uploadAvatarToIpfs/IPFSUploadError.ts
xtyrrell, line 15:
```suggestion
```

xtyrrell: You're right; I reworked this function to do that 👌

xtyrrell:
Thanks for the review :) Cool!

> - uploaded files will not be pinned:
>     @puffo is working another PR that's reworking the ipfsFiles table to be the source of truth for all content. once that's completed, updating the IPFS pinner processor to target that table as you suggest sounds good. can maybe wait for that work to be completed then before merging this PR

Is that https://github.com/spinamp/spindexer/pull/144, which I see is merged?

> - remove avatarIPFSHash column:
>     the above PR is going to be changing the primary key on ipfsFiles so that the cid/hash is actually the primary key, so we'll still need this column as the foreign key

👍 

> - use error processor for retries
>     yeah may make sense to update this in this PR too

Could we do it in a follow-up PR so we can keep this PR from drifting from main? If you'd prefer it's done in this PR happy to do it here though 👍 

> - update the view to join on this table
>     we're gonna be re-working the way artistProfiles get merged together into a single artist record, so i think can hold off on this for now as that future PR should handle merging the avatar url up to the artist record properly

👍

